Issue #11, opened 2021-12-14
Add more tests/examples


lukehoban:
Luke:
- [ ] Clustering (use a shell script instead of inline command)
- [x] Invoking the lambda using `aws lambda invoke` (may require env vars for credentials, etc.)
- [x] curl for managing external SaaS (GitHub labels?)

Laura:
- [x] Provisioner in examples repo
- [ ] Ansible? (Or Chef, etc.)
- [x] `kubectl` command?

Ian:
- [ ] Templating?  (Assets/archives)

nimbinatus:
The remote Ansible example has been breaking in my builds due to either an error with GCP or an error with Ansible itself. There's a very simple local command example I've pushed up, but I'd like to see the remote one working before we call it a real example.

lukehoban:
Closing this out - as we now have a good collection of examples here.  More are always welcome though!

PR #17, opened 2021-12-15
Fix up replacement semantics

### LINKED ISSUE #9: Don't replace on changes to `create`/`delete`
As discussed in https://github.com/pulumi/pulumi-package-command/issues/1#issuecomment-993049181, it is not ideal for the `create` or `delete` properties to force replacement, as these may need to change after a command has been provisioned, but this should not necessarily force the command to re-run.  We should make these not force replacement by default, and then allow users to use `replaceOnChanges` if they desire to actually opt-in to replacement semantics.

lukehoban:
Don't replace on `create`/`delete` changes.  Instead, introduce a new `triggers` property that can be used to indicate dependencies that force replacement.  (Similar to `keepers` in the `random` package or `triggers` in the Terraform `null_resource` provider).

This keeps `connection` as forcing replacement.  It is possible that should be tweaked to not always force a replacement or only force a replacement on `host` changing, but in general if the connection changes, it means the target of the command has changed, which should require re-running.

Also removes the unused `update` property from the schema.

Adds some richer testing to confirm replacement behaviour.

Fixes #9.
Part of #11.


examples/simple/replace/extras.ts
```diff
@@ -0,0 +1,2 @@
+export const len = 20;
+export const fail = false;
```
iwahbe, line 2: Can we add a newline here?

iwahbe (approved):
Besides one nit, LGTM

examples/simple/index.ts
```diff
@@ -1,24 +1,39 @@
...
+    replaceOnChanges: [pw.result],
 }, { deleteBeforeReplace: true });
 
-// const fail = new local.Command("fail", {
-//     create: `echo "couldn't do what I wanted..." && false`,
-// });
+// Manage an external artifact which is created after a resource is created, 
+// deleted before a resource is destroyed, and recreated when the resource is 
+// replaced.  This could also register/deregister the resource with an external 
+// registration or otehr remote API instead of just writing to local disk.
```
nimbinatus, line 26:
```suggestion
// registration or other remote API instead of just writing to local disk.
```

nimbinatus (approved):
tiny nit, but LGTM

PR #21, opened 2021-12-16
Add Remote FileCopy

### LINKED ISSUE #6: Add support for RemoteFile (CopyFile)
We currently support local command and remote commands. The other use case we may need is copying a local file to a local machine. This can technically be accomplished in awkward forms using the other two, but for simple support, may make sense to support directly.

iwahbe:
Adds a resource that represents moving a file from a local to a remote host. 

Fixes #6


provider/cmd/pulumi-resource-command/schema.json
```diff
@@ -180,6 +182,47 @@
...
+        },
+        "command:remote:CopyFile": {
+            "description": "Copy a local file to a remote host.",
+            "inputProperties": {
+                "connection": {
+                    "description": "The parameters with which to connect to the remote host.",
+                    "$ref": "#/types/command:remote:Connection"
+                },
+                "localPath": {
+                    "description": "The path of the file to be copied.",
```
lukehoban, line 194: Can/should we support copying a folder (recursively)?

provider/pkg/provider/provider.go
```diff
@@ -95,19 +106,19 @@ func (k *commandProvider) StreamInvoke(req *pulumirpc.InvokeRequest, server pulu
...
 func (k *commandProvider) Check(ctx context.Context, req *pulumirpc.CheckRequest) (*pulumirpc.CheckResponse, error) {
 	urn := resource.URN(req.GetUrn())
-	ty := urn.Type()
-	if ty != "command:local:Command" && ty != "command:remote:Command" {
-		return nil, fmt.Errorf("unknown resource type %q", ty)
+	err := check(urn)
+	if err == nil {
+		return &pulumirpc.CheckResponse{Inputs: req.News, Failures: nil}, nil
+	} else {
+		return nil, err
```
lukehoban, line 113: Nit for style consistency - replace above with:

```go
if err := check(urn); err != nil {
    return nil, err
}
return &pulumirpc.CheckResponse{Inputs: req.News, Failures: nil}, nil
```

provider/pkg/provider/provider.go
```diff
@@ -95,19 +106,19 @@ func (k *commandProvider) StreamInvoke(req *pulumirpc.InvokeRequest, server pulu
...
 	}
-	return &pulumirpc.CheckResponse{Inputs: req.News, Failures: nil}, nil
 }
 
 // Diff checks what impacts a hypothetical update will have on the resource's properties.
 func (k *commandProvider) Diff(ctx context.Context, req *pulumirpc.DiffRequest) (*pulumirpc.DiffResponse, error) {
```
lukehoban, line 118: I think we'll need to update Diff to handle the new properties - and decide on replacement semantics for `localPath` and `remotePath`.

provider/cmd/pulumi-resource-command/schema.json
```diff
@@ -8,7 +8,8 @@
             "properties": {
                 "user": {
                     "type": "string",
-                    "descripton": "The user that we should use for the connection. Defaults to `root`."
+                    "descripton": "The user that we should use for the connection.",
+                    "default": "root"
```
lukehoban, line 12: I believe with `default` that the `remoteconnection` struct can now type these as non-pointer types, which might clean up some implementation code?

provider/pkg/provider/remotecommand.go
```diff
@@ -39,6 +39,56 @@ type remoteconnection struct {
...
+}
+
+// Dial a ssh client connection from a ssh client configuration, retrying as necessary.
+func (con remoteconnection) Dial(ctx context.Context, config *ssh.ClientConfig) (*ssh.Client, error) {
+	var client *ssh.Client
+	var err error
+	_, _, err = retry.Until(ctx, retry.Acceptor{
+		Accept: func(try int, nextRetryTime time.Duration) (bool, interface{}, error) {
+			client, err = ssh.Dial("tcp",
+				net.JoinHostPort(con.Host, fmt.Sprintf("%d", con.Port)),
```
lukehoban, line 75: I believe you need `*con.Port`.  This is causing the test failure in CI.

provider/pkg/provider/remotefilecopy.go
```diff
@@ -0,0 +1,78 @@
...
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+package provider
+
+import (
+	"context"
+	"os"
+
+	"github.com/pkg/sftp"
```
lukehoban, line 21: Curious - as I'm not certain myself yet - but was there a reason to choose sftp over scp for this?

provider/pkg/provider/remotefilecopy.go
```diff
@@ -0,0 +1,78 @@
...
+	}
+	if err := inner(); err != nil {
+		return "", err
+	}
+	return resource.NewUniqueHex("", 8, 0)
+
+}
+
+func (c *remotefilecopy) RunDelete(ctx context.Context, host *provider.HostClient, urn resource.URN) error {
+	host.Log(ctx, diag.Info, urn, "Deleting remotefilecopy")
```
lukehoban, line 76: I expect we don't need this?

(though we may benefit from debug level diagnostics in quite a few more places for future debuggability)

provider/pkg/provider/remotefilecopy.go
```diff
@@ -0,0 +1,78 @@
...
+			return err
+		}
+
+		_, err = dst.ReadFrom(src)
+		return err
+	}
+	if err := inner(); err != nil {
+		return "", err
+	}
+	return resource.NewUniqueHex("", 8, 0)
```
lukehoban, line 71: Any opportunity to add something more semantically meaningful here?  I could image deriving a prefix from the remotePath potentially?  But maybe more awkward than helpful?

lukehoban (commented):
A few comments from a first pass through.

provider/cmd/pulumi-resource-command/schema.json
iwahbe: That is a non-breaking change, so my thought was we would add it in a later PR.

lukehoban: Yep - makes sense - perhaps open an issue to track this future addition.

iwahbe: https://github.com/pulumi/pulumi-command/issues/23

nimbinatus (approved):
LGTM! I don't have any specific comments about documentation, either; everything looks nice and clean.

provider/pkg/provider/remotefilecopy.go
iwahbe: It would still need a random component because we might have an otherwise identical `CopyFile` with a different `connection`. I don't think there is much benefit. I think the same argument applies to `remote:Command`.

lukehoban: Yeah - agreed. I was wondering similar for remote Command, but there's even less sensible to use there. Sticking with the random hex sounds good.

Issue #4, opened 2021-12-13
Port  pulumi/examples provisioners examples to use this package


lukehoban:
* [x] https://github.com/pulumi/examples/tree/master/aws-py-ec2-provisioners
* [x] https://github.com/pulumi/examples/tree/master/aws-ts-ec2-provisioners
* [x] https://github.com/pulumi/examples/tree/master/classic-azure-ts-vm-provisioners

Note that we might need https://github.com/pulumi/pulumi-package-command/issues/6 as part of this.

PR #24, opened 2021-12-16
Add examples

nimbinatus:
Adds in:

- aws ec2 py example
- gcp compute py example
- simple local Ansible sample to showcase local.Command

ref: #4, #11


nimbinatus:
More coming after dinner :)

lukehoban (approved):
LGTM!

Going to merge this now, will follow up with adding these into the test harness.

PR #25, opened 2021-12-17
Allow `CopyFile:localPath` to accept directory paths, `Assets` and `Archives`

iwahbe:
This allows `localPath` to operate on directories. It also allows it to accept Pulumi types (`Asset`, `Archive`).


examples/remote_exec/index.ts
```diff
@@ -0,0 +1,75 @@
...
+    connection,
+    localPath: new asset.FileAsset("./program.py"),
+    remotePath: "program.py",
+});
+
+const zipDataPath = "/tmp/data.zip";
+
+const zipData = new local.Command("zip", {
+    create: `zip ${zipDataPath} data/*`,
+    delete: `rm ${zipDataPath}`,
```
lukehoban, line 58: Does this fail if the path doesn't exist?  Should it be `|| true` to allow the delete to succeed if the path doesn't exist (presuming it's run on a machine that didn't run the create?

examples/remote_exec/index.ts
```diff
@@ -0,0 +1,75 @@
...
+});
+
+const data = new remote.CopyFile("data", {
+    connection,
+    localPath: zipData.stdout.apply((_: string) => new asset.FileArchive("/tmp/data.zip")),
```
lukehoban, line 63: Why is there the need for the `apply` here?  Could this be accomplished with a `dependsOn`?

Also - minor:
```suggestion
    localPath: zipData.stdout.apply((_: string) => new asset.FileArchive(zipDataPath)),
```

examples/remote_exec/index.ts
```diff
@@ -0,0 +1,75 @@
...
+};
+
+const checkPython = new remote.Command("versionCheck", {
+    connection,
+    create: "python --version",
+});
+
+const program = new remote.CopyFile("program", {
+    connection,
+    localPath: new asset.FileAsset("./program.py"),
```
lukehoban, line 50: The benefit here I presume is that the copy-file replaces if the contents of `program.py` change?  Is that right?

provider/cmd/pulumi-resource-command/schema.json
```diff
@@ -191,8 +191,8 @@
                     "$ref": "#/types/command:remote:Connection"
                 },
                 "localPath": {
-                    "description": "The path of the file to be copied.",
-                    "type": "string"
+                    "description": "The file/folder to be copied.\nIf the input is an Asset, it will be interperted as the contents of a file.\nIf the input is an Archive, it will be interperted as the contents of a folder.\nIf the input is a string, it will be interpreted as the path to a file or folder.\nBecause assets and archives are understood by the Pulumi model, they use should be\nprefered to raw paths.",
```
lukehoban, line 194: The main reason an Asset or Archive could be more useful than a path string in general is that the asset/archive computes the hash, which allows identifying whether the file contents have changed, and thus triggering an update/replace.  Is that accomplished here?  If so, might be good to call that out as the reason for preferring Asset/Archive?

provider/cmd/pulumi-resource-command/schema.json
```diff
@@ -191,8 +191,8 @@
...
+                    "$ref": "pulumi.json#/Any"
```
lukehoban, line 195: It might be nicer to offer three separate properties?  `localPath`, `localAsset`, `localArchive`?  Then each could be strongly typed.  They would all have to be optional, and to be mutually exclusive.  

Not 100% on this, but the weak typing that results here is a little unfortunate.

lukehoban (commented):
Biggest question is whether this causes update/replace when the asset/archive contents change (but the path does not).  That is typically the reason why Asset/Archive is used, but I don't see anything explicitly ensuring that will happen here.

examples/remote_exec/index.ts
iwahbe: It should, yes. I'll make a note in the example code and the schema comment.

provider/cmd/pulumi-resource-command/schema.json
iwahbe: I 100% agree that the weak typing is unfortunate. To me, allowing 3 mutually exclusive fields is also weak typing. We allow the type system to accept invalid types. What I should have used (having just remembered this) is `"oneOf"`. This allows us to express a union type. I'll push an update with this.

provider/cmd/pulumi-resource-command/schema.json
```diff
@@ -191,8 +191,12 @@
                     "$ref": "#/types/command:remote:Connection"
                 },
                 "localPath": {
-                    "description": "The path of the file to be copied.",
-                    "type": "string"
+                    "description": "The file/folder to be copied.\nIf the input is an Asset, it will be interpreted as the contents of a file.\nIf the input is an Archive, it will be interpreted as the contents of a folder.\nIf the input is a string, it will be interpreted as the path to a file or folder.\nAsset's and Archives change depending on their contents. This means that a change\nto the underlying file/folder will result in a \"replace\" operation for the Copy.",
```
nimbinatus, line 194: Grammar fix.

```suggestion
                    "description": "The file/folder to be copied.\nIf the input is an Asset, it will be interpreted as the contents of a file.\nIf the input is an Archive, it will be interpreted as the contents of a folder.\nIf the input is a string, it will be interpreted as the path to a file or folder.\nAssets and Archives change depending on their contents. This means that a change\nto the underlying file/folder will result in a \"replace\" operation for the Copy.",
```

nimbinatus: Question: The capitalization here is slightly different than I would expect, and I want to ensure we're clear for someone who might not know the system as well to understand the docs. Assets and Archives are, in this case, Pulumi concepts, correct? It's noticeable because we are not overwriting the general sense of the term `string` in comparison. And Copy, as well? I would refer to the action in full in the latter case (`CopyFile`) to be clear:

`Object` isn't exactly the right term as I know we use languages that are not object-oriented, but hopefully this suggestion makes clear what I'm trying to do:

```suggestion
                    "description": "The file/folder to be copied.\nIf the input is an Asset object, it will be interpreted as the contents of a file.\nIf the input is an Archive object, it will be interpreted as the contents of a folder.\nIf the input is a string, it will be interpreted as the path to a file or folder.\nAsset objects and Archive objects change depending on their contents. This means that a change\nto the underlying file/folder will result in a \"replace\" operation for the CopyFile call.",
```

provider/pkg/provider/provider.go
```diff
@@ -244,10 +251,48 @@ func (k *commandProvider) Read(ctx context.Context, req *pulumirpc.ReadRequest)
 		return nil, err
 	}
 
+	// This transformation is necessary here because "localPath" is a union of assets and archives
+	// (and a string). Asset's and Archives are not serialized as valid property maps by default.
```
nimbinatus, line 255:
```suggestion
	// (and a string). Assets and Archives are not serialized as valid property maps by default.
```

examples/remote_exec/program.py
```diff
@@ -0,0 +1,16 @@
+#!/usr/bin/env python3
+
+import sys
+import os
```
nimbinatus, line 4: Some notes (non-blockers):

PEP 8 would have us order standard library calls alphabetically, and I generally try to ensure every example we put out follows all standard style guidelines to avoid people complaining about trivial things. Same with spaces on line 10 around the plus signs (and, to be honest, I'd probably use an f-string there for readability, but that's more of a preference than a style note).

nimbinatus (changes requested):
One grammar adjustment in the docs, a couple suggestions, and a non-blocking note.

examples/remote_exec/program.py
iwahbe: I'll fix the imports and the plug signs (and run the whole thing through `pylint`). I'm avoiding using f-string because ec2 instances have python(2) but not python3. I don't want to make the example run even longer by installing python3.

provider/cmd/pulumi-resource-command/schema.json
iwahbe: I've heard them referred to as "types", and I think that language is universal. Thanks for pointing this out!

PR #27, opened 2021-12-17
Add an example which uses cURL to manage GitHub labels

lukehoban:
Also shows using Command within a component to create a nice typed interface.  This includes parsing stdout (JSON) to extract strongly typed results.

Also shows that secrets compose into this model.

Also shows offloading commands to local scripts.

Part of #11.


nimbinatus (approved):
+1, LGTM

PR #34, opened 2022-01-10
remove whitespace from logo

susanev:
im removing the whitespace from the logo so it grows to the correct size in the registry


susanev:
@lukehoban i dont have access to this repo so i used a fork; also seems like i cant add you as a reviewer so tagging you here instead.

lukehoban (commented):
Thanks!

PR #36, opened 2022-01-19
Update Makefile

flashspys:
The tests folder was renamed to examples. I made this change also in the makefile


stack72:
Thanks so much for this @flashspys

stack72 (approved):

PR #38, opened 2022-01-24
Update README to fix incorrect Go import path

itay:
The Go example had an import at `v3/go/` instead of just `go/`


viveklak (approved):
Thanks @itay !

Issue #37, opened 2022-01-24
local.Command not re-running on changing inputs


itay:
## Hello!

- Vote on this issue by adding a 👍 reaction
- To contribute a fix for this issue, leave a comment (and link to your pull request, if you've opened one already)

## Issue details

I'm trying a relatively simple usage (similar to the GithubLabels example), but it's not really working as I expected. This is using Go and v0.0.3 of local.Command.

Here is my `main.go`:

```
package main

import (
	"github.com/pulumi/pulumi-command/sdk/go/command/local"
	"github.com/pulumi/pulumi/sdk/v3/go/pulumi"
)

func main() {
	pulumi.Run(func(ctx *pulumi.Context) error {

		_, err := local.NewCommand(ctx, "test", &local.CommandArgs{
			Create: pulumi.String("./create.sh hello"),
			Delete: pulumi.String("./destroy.sh goodbye"),
			Environment: pulumi.StringMap{
				"foo": pulumi.String("bar"),
			},
		})
		if err != nil {
			return err
		}

		return nil
	})
}
```

Here is my `create.sh` and `destroy.sh`:

```
#!/bin/bash

echo "hello: $1" >> /tmp/pulumi_log.txt
```

```
#!/bin/bash

echo "goodbye: $1" >> /tmp/pulumi_log.txt
```

When I run `pulumi up` the first time, the resource gets created and I see the right entry in the log file. Similarly, if I explicitly run `pulumi destroy`, I see the right entry added to the log file.

However, if I were to change the value (e.g. changing `"bar"` to `"bar2"`), and I run `pulumi up`, it tells me that `environment` is updated, but the script never executes again (and no corresponding entry in the log).

Not sure if I'm doing something wrong or have the wrong expectations.

### Steps to reproduce

See above.

Expected: Script re-runs when the inputs change.
Actual: Script doesn't re-run.

viveklak:
This seems by design as per https://github.com/pulumi/pulumi-command/blob/master/provider/pkg/provider/provider.go#L278. Once could trigger an update by specifying a replaceOnChanges clause for environment etc. cc @lukehoban if this seems worth doing by default or if there is a good use case to consider offering an update hook in addition.

itay:
Per the discussion in Slack: https://pulumi-community.slack.com/archives/C84L4E3N1/p1643048823205900

[...]

PR #57, opened 2022-05-11
Add update fields for local and remote command

jkisk:
Add an optional update field to the local and remote command resources that will run on update if supplied.  If this field is omitted an update operation will re-run the create field command.

Resolves: #37

### DIFF
--- a/provider/pkg/provider/command.go
+++ b/provider/pkg/provider/command.go
@@ -38,6 +38,8 @@ type command struct {
 	Triggers    *[]interface{}     `pulumi:"triggers,optional"`
 	Create      string             `pulumi:"create"`
 	Delete      *string            `pulumi:"delete,optional"`
+	// Optional, if empty will run Create again
+	Update      *string			   `pulumi:"update,optional"`
 	Stdin       *string            `pulumi:"stdin,optional"`
 
 	// Output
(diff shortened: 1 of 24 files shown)


provider/pkg/provider/command.go
```diff
@@ -54,6 +56,20 @@ func (c *command) RunCreate(ctx context.Context, host *provider.HostClient, urn
 	return id, err
 }
 
+//
+func (c *command) RunUpdate(ctx context.Context, host *provider.HostClient, urn resource.URN) (string, error) {
+	if c.Update != nil {
+		stdout, stderr, id, err := c.run(ctx, *c.Update, host, urn)
```
stack72, line 62: Should we dereference Update or not (we don't with Create below)

provider/pkg/provider/remotecommand.go
```diff
@@ -124,6 +125,19 @@ func (c *remotecommand) RunDelete(ctx context.Context, host *provider.HostClient
 	return err
 }
 
+func (c *remotecommand) RunUpdate(ctx context.Context, host *provider.HostClient, urn resource.URN) (string, error) {
+	if c.Update != nil {
+		stdout, stderr, id, err := c.run(ctx, *c.Update, host, urn)
```
stack72, line 130: Same question as before

stack72:
Would be really good to add an update test for this functionality

provider/pkg/provider/command.go
jkisk: We should, because update is *string, similar to the delete field, where create is a string.

stack72 (approved):
Merge on a green PR run

rbm-radius:
Yay! Thank you!

PR #60, opened 2022-05-20
Update x/crypto to support RSA SHA-2 client authentication

blampe:
`ssh-keygen` by default now generates RSA SHA-2 private keys (of the form `BEGIN OPENSSH PRIVATE KEY`). 

Currently, executing a `remote.Command` with one of these keys will fail with

```
unable to connect: ssh: handshake failed: ssh: unable to authenticate, attempted methods [none publickey], no supported methods remain
```

This change updates x/crypto to pull in https://github.com/golang/crypto/commit/5d542ad81a58c89581d596f49d0ba5d435481bcf, which implements support for RSA SHA-2 client authentication. I've confirmed this works as expected locally.


stack72:
Thanks so much for this @blampe!!

PR #66, opened 2022-05-31
Use Go 1.17

jkisk:


stack72 (commented):
don't we need to re-run go mod tidy on examples/go.mo?

stack72 (approved):

PR #73, opened 2022-06-14
[internal] ensure sdk tag for go sdk pushed as part of release

stack72:


jkisk (commented):
Should we add a conditional to only do this `if: success()`?

Issue #112, opened 2022-08-02
Java support


myhau:
## Hello!

- Vote on this issue by adding a 👍 reaction
- If you want to implement this feature, comment to let us know (we'll work with you on design, scheduling, etc.)

## Issue details

This is about about adding support for Java — at least maven artifacts and docs with examples should be generated.

PR #119, opened 2022-08-17
Add support for Java to the published SDK

stack72:
Fixes: #112


jkisk (approved):
Looks good!  All the go changes seem to be the comment formatting from the update to go 1.19 and the workflow changes are as expected.

PR #122, opened 2022-08-23
Prevent against panic concurrent writes in addContext

stack72:
Fixes: #88


jkisk (approved):

provider/pkg/provider/provider.go
```diff
@@ -465,7 +471,9 @@ func (k *commandProvider) Cancel(context.Context, *pbempty.Empty) (*pbempty.Empt
 	return &pbempty.Empty{}, nil
 }
 
-func (k *commandProvider) addContext(c context.Context) context.Context {
+func (k *commandProvider) addContext(c context.Context, mutex *sync.Mutex) context.Context {
+	mutex.Lock()
+	defer mutex.Unlock()
```
danielrbradley, line 476: I might be miss-understanding the go-specific mutex implementation, but it appears that we're creating a new mutex each time we're calling `addContext` at which point none of these would block each other from modifying `k.cancelFuncs` at the time time. I would have expected to see the mutex object initialized on the `commandProvider` (`k`) struct so we have a single instance which all threads would lock against.

stack72: I've changed this around now

provider/pkg/provider/provider.go
```diff
@@ -466,6 +469,8 @@ func (k *commandProvider) Cancel(context.Context, *pbempty.Empty) (*pbempty.Empt
 }
 
 func (k *commandProvider) addContext(c context.Context) context.Context {
+	k.providerMutex.Lock()
+	defer k.providerMutex.Unlock()
```
danielrbradley, line 473: We should also add this wherever there's operations on the map - so also in the remove methon and cancel flow.

danielrbradley (changes requested):

provider/pkg/provider/provider.go
danielrbradley: Have added in 707450d

danielrbradley (approved):

provider/pkg/provider/provider.go
stack72: Good addition

PR #126, opened 2022-09-07
Fix properties overwrite on refresh

### LINKED ISSUE #94: Stdout (and stderr) are unexpectedly deleted on a refresh operation
### What happened?

We have a script we're running using the command resource that outputs data about the objects that it is creating. We then consume and parse the stdout output to pass this information into dependent resources. However when we run a refresh on the stack, the stdout and stderr output values are cleared and then because the Command resource already exists the output is never available again unless we force the resource to be recreated.

### Steps to reproduce

Create a command resource that outputs a string value to stdout.
Create another resource that uses stdout from the command resource.
Run a refresh.
The dependent resource will no longer have a value from stdout.

### Expected Behavior

Either the refresh should not delete stdout and stderr, or the Command resource needs to have a "Get" command to output the current value of the resource so that refresh will update stdout and stderr instead of deleting them.

### Actual Behavior

stdout and stderr are deleted on pulumi refresh

### Versions used

The pulumi-command version used was 0.3.2, I don't have access to the project at the moment to get the other versions. Let me know if I need to provide them and I'll update when I can.

### Additional context

_No response_

### Contributing

brumhard:
Fixes #94 
I think this was just a simple typo. I validated that it works as expected for me after this change.

### DIFF
--- a/provider/pkg/provider/provider.go
+++ b/provider/pkg/provider/provider.go
@@ -315,7 +315,7 @@ func (k *commandProvider) Read(ctx context.Context, req *pulumirpc.ReadRequest) 
 	return &pulumirpc.ReadResponse{
 		Id:         req.GetId(),
 		Inputs:     req.GetInputs(),
-		Properties: req.GetInputs(),
+		Properties: req.GetProperties(),
 	}, nil
 }
 


brumhard:
There's no changelog file right?

brumhard:
@iwahbe can you maybe have a look at it? 🙂

iwahbe:
Yep, this just looks like a typo. Thanks for posting the fix.

iwahbe (approved):

brumhard:
@iwahbe thank you 👍🏻 Are there any plans on when the next version is released?

PR #128, opened 2022-09-10
support for remote command with ssh-agent

### LINKED ISSUE #96: Support using SSH agent for authentication
## Hello!

## Issue details

My instance allows SSH access via a pre-existing public key, which I have added to my locally running SSH agent (with `ssh-add`), so I can just `ssh example.com` without specifying or configuring a private key.

`remote.Command` does not detect that an SSH agent is available, so it fails with the following error:
```
  command:remote:CopyFile (master.sh):
    error: ssh: handshake failed: ssh: unable to authenticate, attempted methods [none], no supported methods remain
```

If the environment variable `$SSH_AUTH_SOCK` is defined (or perhaps via an input option), the connection should create a new SSH client using the [agent package](https://pkg.go.dev/golang.org/x/crypto@v0.0.0-20220622213112-05595931fe9d/ssh/agent#NewClient), and use this to respond, as described [here](https://blog.ralch.com/articles/golang-ssh-connection/#ssh-agent).

protometa:
fixes #96

The change is very simple but I'd definitely need some advice on how to write a test for this kind of thing.

### DIFF
--- a/provider/pkg/provider/remotecommand.go
+++ b/provider/pkg/provider/remotecommand.go
@@ -25,6 +25,7 @@ import (
 	"time"
 
 	"golang.org/x/crypto/ssh"
+	"golang.org/x/crypto/ssh/agent"
 
 	"github.com/pulumi/pulumi/pkg/v3/resource/provider"
 	"github.com/pulumi/pulumi/sdk/v3/go/common/diag"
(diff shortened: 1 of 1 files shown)


dixler:
Hi. It seems that #96 is closed so I will close this PR. Thanks for submitting this.

PR #118, opened 2022-08-17
Spike for the go-provider-pkg library

iwahbe:

Issue #111, opened 2022-07-31
`Inputs.ConnectionArgs.PrivateKey` leaks private keys in plain text by default


Smaug123:
### What happened?

The easiest way to use the `PrivateKey` feature of `ConnectionArgs` will leak your private key into the `connection` property of the resulting `Command`'s resource (specifically the `privateKey` of the JSON blob of that value).

### Steps to reproduce

F# to construct a `Command` which will leak your private key if you run it:

```fsharp
module PulumiWebServer.Thing

open Pulumi.Command.Remote

let conn =
    let inputArgs = Inputs.ConnectionArgs ()

    inputArgs.Host <- "192.168.0.1"
    inputArgs.Port <- 22.0
    inputArgs.User <- "root"
    inputArgs.PrivateKey <- System.IO.File.ReadAllText "~/.ssh/id_ed25519"
    inputArgs

let bad_command () =
    let args = CommandArgs ()
    args.Connection <- conn
    args.Create <- "echo hello"

    Command ("oh-no", args)
```

### Expected Behavior

The `connection` property to contain a hash of my private key.

### Actual Behavior

The `connection` property contained my private key in plain text.

### Versions used

```
pulumi about
CLI          
Version      3.35.2
Go Version   go1.17.11
Go Compiler  gc

Plugins
NAME          VERSION
cloudflare    4.8.0
command       0.4.1
digitalocean  4.14.0
dotnet        unknown

Host     
OS       darwin
Version  12.4
Arch     arm64

This project is written in dotnet: executable='/nix/store/ypjjp011l3crp0m3wmp1xbwxn5q5xpdr-dotnet-sdk-6.0.301/bin/dotnet' version='6.0.301'

Current Stack: dev

TYPE                                URN
pulumi:pulumi:Stack                 urn:pulumi:dev::PulumiWebServer::pulumi:pulumi:Stack::PulumiWebServer-dev
pulumi:providers:digitalocean       urn:pulumi:dev::PulumiWebServer::pulumi:providers:digitalocean::default_4_14_0
digitalocean:index/sshKey:SshKey    urn:pulumi:dev::PulumiWebServer::digitalocean:index/sshKey:SshKey::default
digitalocean:index/droplet:Droplet  urn:pulumi:dev::PulumiWebServer::digitalocean:index/droplet:Droplet::nixos-server
pulumi:providers:cloudflare         urn:pulumi:dev::PulumiWebServer::pulumi:providers:cloudflare::default_4_8_0
[...]

Issue #22, opened 2021-12-16
Add support for passphrase encrypted private keys in `connection`


lukehoban:

JamesMilneFilmlight:
Just ran into this one, trying to migrate from the old ' dynamic provisioners' example to the new Pulumi Command package.

danielbraun89:
This is a real blocker for me as well

PR #156, opened 2022-11-07
Go SDK provider

dixler:
Adopts changes from https://github.com/pulumi/pulumi-command/pull/118 and merges changes from master branch.

Fixes: https://github.com/pulumi/home/issues/2311

TODO:
1. - [x] Port changes from
      - [x] https://github.com/pulumi/pulumi-command/issues/142
      - [x] https://github.com/pulumi/pulumi-command/issues/111
      - [x] https://github.com/pulumi/pulumi-command/issues/22
      - [x] https://github.com/pulumi/pulumi-command/issues/96
2. - [x] Implement WireDependencies to implement **proper** secrets handling.
3. - [x] verify schema. Notes on differences below.
     - the command:local:Run reuses arguments from command:local:Command and inherits `AssetPaths` and `ArchivePaths`
     - Some descriptions were enhanced in https://github.com/pulumi/pulumi-command/pull/118 and those changes are adopted.
     - Some fields were marked as replace on changes to pass `TestSimple`
     - pulumi-go-provider SDK
       -  omits empty list fields
       - adds empty Object fields 
5. - [x] Break resource/function type definitions and annotations into smaller files. (this provider is meant to be an example so the files should be smaller and focused.)
5. - [x] Refactor
      - [x] Rename args/state to output/input
      - [x] delete language specific codegen stuff
5. - [x] Fix broken tests
```
Diagnostics:
    command:local:Command (pwd):
      error: rm: cannot remove 'aW<iTCfmT&dog.txt': No such file or directory
      error: exit status 1
      error: rm: cannot remove 'aW<iTCfmT&dog.txt': No such file or directory
      error: exit status 1
  
    pulumi:pulumi:Stack (command-simple-with-update-p-it-fv-az290-6-simple-wit-96bf14f7):
      error: update failed
```
7. - [x] Comment(I'm going to extensively comment 1 of each type rather than everything obsessively.)
      - [x] Schema
      - [x] Resources
      - [x] Functions
      - [x] Language Map


provider/pkg/provider/remote/command.go
```diff
@@ -0,0 +1,72 @@
...
+func (c *Command) Annotate(a infer.Annotator) {
+	a.Describe(&c, `A command to run on a remote host.
+The connection is established via ssh.`)
+}
+
+// The arguments for a remote Command resource
+type CommandInputs struct {
+	// these field annotations allow you to
+	// pulumi:"connection" specifies the name of the field in the schema
+	// pulumi:"optional" specifies that a field is optional. This must be optional.
```
iwahbe, line 32: I'm not sure what "This must be optional." means. If the type is not optional, it gets a zero value if not filled in.

provider/pkg/provider/remote/command.go
```diff
@@ -0,0 +1,72 @@
...
+	// provider:"replaceOnChanges" specifies that a resource will be marked replaceOnChanges.
+	// provider:"secret" specifies that a resource will be marked replaceOnChanges.
```
iwahbe, line 34:
```suggestion
	// provider:"secret" specifies that a field should be marked secret.
```

provider/pkg/provider/remote/command.go
```diff
@@ -0,0 +1,72 @@
...
+	a.Describe(&c, `A command to run on a remote host.
+The connection is established via ssh.`)
+}
+
+// The arguments for a remote Command resource
+type CommandInputs struct {
+	// these field annotations allow you to
+	// pulumi:"connection" specifies the name of the field in the schema
+	// pulumi:"optional" specifies that a field is optional. This must be optional.
+	// provider:"replaceOnChanges" specifies that a resource will be marked replaceOnChanges.
```
iwahbe, line 33:
```suggestion
	// provider:"replaceOnChanges" specifies that the resource will be replaced if the field changes.
```

provider/pkg/provider/remote/copyfileController.go
```diff
@@ -0,0 +1,83 @@
...
+// These are not required. They indicate to Go that Command implements the following interfaces.
+// If the function signature doesn't match or isn't implemented, we get nice compile time errors in this file.
+var _ = (infer.CustomResource[CopyFileInputs, CopyFileOutputs])((*CopyFile)(nil))
+var _ = (infer.ExplicitDependencies[CopyFileInputs, CopyFileOutputs])((*CopyFile)(nil))
+
+func (r *CopyFile) WireDependencies(f infer.FieldSelector, args *CopyFileInputs, state *CopyFileOutputs) {
+	f.OutputField(&state.CopyFileInputs.Connection).DependsOn(f.InputField(&args.Connection))
+	f.OutputField(&state.CopyFileInputs.Triggers).DependsOn(f.InputField(&args.Triggers))
+	f.OutputField(&state.CopyFileInputs.LocalPath).DependsOn(f.InputField(&args.LocalPath))
+	f.OutputField(&state.CopyFileInputs.RemotePath).DependsOn(f.InputField(&args.RemotePath))
```
iwahbe, line 37: Do you think there should be an `DependOnMatchingInputs` method on `infer.FieldSelector` to do this without all the boilerplate? The semantics would be `state.A` depends on `args.B` iff the pulumi name of `args.B` matches the pulumi name of `state.A`.

I suspect this will be a common pattern. My only concern is discoverability. @dixler what do you think?

iwahbe (approved):
The overall design looks good to me. I left a couple of comments, but overall it looks good.

provider/pkg/provider/remote/copyfileController.go
dixler: IMO I think it's better to force the author to be a little more explicit rather than depend on implicit/magic default behavior. If someone explicitly asks for it in the future, we can add it.

provider/pkg/provider/local/commandController.go
```diff
@@ -0,0 +1,144 @@
...
+	}
+	if input.Create == nil {
+		return id, state, nil
+	}
+	cmd := *input.Create
+	state.Stdout, state.Stderr, err = state.run(ctx, cmd)
+	return id, state, err
+}
+
+// WireDependencies is relevant to secrets handling. This method indicates the what Inputs
```
aq17, line 52:
```suggestion
// WireDependencies is relevant to secrets handling. This method indicates which Inputs
```

provider/pkg/provider/remote/copyfileController.go
iwahbe: Sounds good. BTW, I wasn't suggesting that this becomes the default behavior, I was suggesting that we add a method (that the user needs to call) to do this.

My concern is that if you add a new field to `CopyFileInputs`, you might forget to add it to this block.

provider/pkg/provider/remote/command.go
iwahbe: I don't think the type needs to be a pointer. Since all types in go have a zero value, I think it can be any type.

PR #169, opened 2022-12-27
Allow `delete` and `update` operations to access `stdout` via `PULUMI_COMMAND_STDOUT`.

### LINKED ISSUE #167: Support accessing stdout state from within `delete`
It is common for the `delete` operation to need to customize it's action based on the results of the `create` operation.  Although the resource's `stdout` is available from within Pulumi, it is not currently available from within the `delete` command.

For example - a user asked how to accomplish this:

```py
mktemp = Command('mktemp', create='mktemp', delete='rm {path}')
```

The `{path}` piece of this doesn't work - and there is no way to know which temp dir to delete without having access to the `stdout` of the `create`.

There are a few ways we could support this:
1.  Process the delete command string as an interpolation, passing in a variable like `stdout` to be (potentially) interpolated.
2. Use the existing support for env vars, and always inject a well-known env var like `PULUMI_COMMAND_STDOUT` into the environment for `delete` operations.

I am slightly more inclined to do (2) as it reuses the same existing approach of relying on environment variables to pass data into the command, instead of inventing a new non-bash notion of interpolation.

lukehoban:
Fixes #167.

### DIFF
--- a/provider/pkg/provider/local/command.go
+++ b/provider/pkg/provider/local/command.go
@@ -58,12 +58,13 @@ func (c *CommandInputs) Annotate(a infer.Annotator) {
 	c.BaseInputs.Annotate(a)
 	a.Describe(&c.Triggers, "Trigger replacements on changes to this input.")
 	a.Describe(&c.Create, "The command to run on create.")
-	a.Describe(&c.Delete, "The command to run on delete. The environment variables PULUMI_COMMAND_STDOUT "+
-		"and PULUMI_COMMAND_STDERR are set to the stdout and stderr properties of the Command resource from "+
-		"previous create or update steps.")
-	a.Describe(&c.Update, "The command to run on update, if empty, create will run again. The environment "+
-		"variables PULUMI_COMMAND_STDOUT and PULUMI_COMMAND_STDERR are set to the stdout and stderr properties "+
-		"of the Command resource from previous create or update steps.")
+	a.Describe(&c.Delete, `The command to run on delete. The environment variables PULUMI_COMMAND_STDOUT
+and PULUMI_COMMAND_STDERR are set to the stdout and stderr properties of the
+Command resource from previous create or update steps.`)
+	a.Describe(&c.Update, `The command to run on update, if empty, create will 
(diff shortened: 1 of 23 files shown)


provider/pkg/provider/local/command.go
```diff
@@ -58,8 +58,12 @@ func (c *CommandInputs) Annotate(a infer.Annotator) {
...
 	a.Describe(&c.Triggers, "Trigger replacements on changes to this input.")
 	a.Describe(&c.Create, "The command to run on create.")
-	a.Describe(&c.Delete, "The command to run on delete.")
-	a.Describe(&c.Update, "The command to run on update, if empty, create will run again.")
+	a.Describe(&c.Delete, "The command to run on delete. The environment variables PULUMI_COMMAND_STDOUT "+
+		"and PULUMI_COMMAND_STDERR are set to the stdout and stderr properties of the Command resource from "+
+		"previous create or update steps.")
+	a.Describe(&c.Update, "The command to run on update, if empty, create will run again. The environment "+
+		"variables PULUMI_COMMAND_STDOUT and PULUMI_COMMAND_STDERR are set to the stdout and stderr properties "+
+		"of the Command resource from previous create or update steps.")
```
iwahbe, line 66: I would insert some newlines in the description. Our codegen doesn't break up long lines in comments, so it's helpful to have them here. 

```suggestion
	a.Describe(&c.Delete, `The command to run on delete. The environment variables PULUMI_COMMAND_STDOUT
	and PULUMI_COMMAND_STDERR are set to the stdout and stderr properties of the
	Command resource from previous create or update steps.`)
	a.Describe(&c.Update, `The command to run on update, if empty, create will 
run again. The environment  variables PULUMI_COMMAND_STDOUT and PULUMI_COMMAND_STDERR 
are set to the stdout and stderr properties of the Command resource from previous 
create or update steps.`)
```

iwahbe (approved):
I appreciate the test! I like the use of env vars as a language agnostic way to add what is effectively additional inputs.

LGTM!

PR #170, opened 2022-12-29
Improve stdout/stderr handling

### LINKED ISSUE #80: Print stdout too when there is a failure
It appears right now that only stderr is emitted to the resource's output stream.  hat could be the right default, to avoid excessive noise in the output, however this makes debugging failures very hard, requiring that you resort to `--logtostderr`. I've tried exporting the `stdout` property as a stack output, but because the resource creation fails, it doesn't get emitted.

At the very least, it seems we should include both stderr and stdout in the output in the case of failure.

I also wonder whether most people will expect the output by default, at least as an ephemeral stream (I know I did, similar to our Docker package). That might be a better default and then we could offer `suppressStdout` and `supressStderr` options.

### LINKED ISSUE #121: A command that writes to stderr gives confusing Pulumi output
Given the following Pulumi program:

```typescript
import * as command from "@pulumi/command";

new command.local.Command("docker-build", {
  create: "echo \"hello world\" >&2"
});
```

I get the following output:

```bash
$ pulumi up -y
Previewing update (dev)

View Live: https://app.pulumi.com/jkodroff/try-docker-command/dev/previews/24f5e5f1-1143-413e-b6d1-736b7f34fefc

     Type                      Name                    Plan
 +   pulumi:pulumi:Stack       try-docker-command-dev  create
 +   └─ command:local:Command  docker-build            create

Resources:
    + 2 to create

Updating (dev)

View Live: https://app.pulumi.com/jkodroff/try-docker-command/dev/updates/27

     Type                      Name                    Status                  Info
 +   pulumi:pulumi:Stack       try-docker-command-dev  created
 +   └─ command:local:Command  docker-build            **creating failed**     1 error

Diagnostics:
  command:local:Command (docker-build):
    error: hello world

Resources:
    + 2 created

Duration: 2s
```

The output is confusing because the stack tree shows an error, but the `Resources:` summary shows success. Pulumi itself also returns 0.

I ran into this because `docker build` outputs to stderr. Per [this issue](https://github.com/moby/buildkit/issues/1186#issuecomment-1090427273), it looks like printing logs to stderr is common in many *nix tools (to my surprise).

If indeed printing progress messages to stderr is common, I think we need to ensure that by default, stderr output on `local.Command` does not display as an error in Pulumi. We may also want to consider an option in the Command resource to keep the existing behavior.

lukehoban:
This overhauls the treatment of stdout and stderr for local.Command and remote.Command.

The new rules are:
1. stdout and stderr are both rendered directly to ephemeral logs always.
2. when the command fails, the resource creation fails, and the returned error message includes the full combined text of stdout+stderr (along with the exit code and the command that was run).
3. the stdout and stderr properties of the resource are still populated with the values of those output streams if the command succeeds.

In the future, we may want to support an option to suppress (1) for cases where the command output if overly verbose, but it is better to include it by default (similar to the docker.Image component).

Fixes #80.
Fixes #121.

### DIFF
--- a/examples/simple-run/index.ts
+++ b/examples/simple-run/index.ts
@@ -2,7 +2,7 @@ import * as local from "@pulumi/command/local";
 import * as random from "@pulumi/random";
 import { interpolate } from "@pulumi/pulumi";
 
-const pw = new random.RandomPassword("pw", { length: 10 });
+const pw = new random.RandomPassword("pw", { length: 10, special: false });
 
 const plainFile = local.runOutput({
     command: `echo "Hello world!" > hello.txt`,
--- a/provider/pkg/provider/local/commandOutputs.go
+++ b/provider/pkg/provider/local/commandOutputs.go
@@ -46,18 +46,13 @@ func (c *CommandOutputs) run(ctx p.Context, command string) (string, string, err
 	}
 	args = append(args, command)
 
-	stdoutr, stdoutw, err := os.Pipe()
-	if err != nil {
-		return "", "", err
-	}
-	stderrr, stderrw, err := os.Pipe()
-	if err != nil {
-		return "", "", err
-	}
+	var err error
+	var stdoutbuf, stderrbuf, stdouterrbuf bytes.Buffer
+	r, w := io.Pipe()
 
 	cmd := exec.CommandContext(ctx, args[0], args[1:]...)
-	cmd.Stdout = stdoutw
-	cmd.Stderr = stderrw
+	cmd.Stdout = io.MultiWriter(&stdoutbuf, &stdouterrbuf, w)
+	cmd.Stderr = io.MultiWriter(&stderrbuf, &stdouterrbuf, w)
 	if c.Dir != nil {
 		cmd.Dir = *c.Dir
 	} else {
@@ -83,24 +78,15 @@ func (c *CommandOutputs) run(ctx p.Context, command string) (string, string, err
 		cmd.Stdin = strings.NewReader(*c.Stdin)
 	}
 
-	var stdoutbuf bytes.Buffer
-	var stderrbuf bytes.Buffer
-	var stdouterrbuf bytes.Buffer
-
-	stdouttee := io.TeeReader(stdoutr, &stdoutbuf)
-	stderrtee := io.TeeReader(stderrr, &stderrbuf)
-	stdouterr := io.MultiReader(stdouttee, stderrtee)
-	stdouterr = io.TeeReader(stdouterr, &stdouterrbuf)
 	stdouterrch := make(chan struct{})
-	go util.CopyOutput(ctx, stdouterr, stdouterrch, diag.Info)
+	go util.CopyOutput(ctx, r, stdouterrch, diag.Info)
 
 	err = cmd.Start()
 	if err == nil {
 		err = cmd.Wait()
 	}
 
-	stdoutw.Close()
-	stderrw.Close()
+	w.Close()
 	<-stdouterrch
 
 	if err != nil {
--- a/provider/pkg/provider/local/commandOutputs.go
+++ b/provider/pkg/provider/local/commandOutputs.go
@@ -48,11 +48,12 @@ func (c *CommandOutputs) run(ctx p.Context, command string) (string, string, err
 
 	var err error
 	var stdoutbuf, stderrbuf, stdouterrbuf bytes.Buffer
+	stdouterrwriter := util.ConcurrentWriter{Writer: &stdouterrbuf}
 	r, w := io.Pipe()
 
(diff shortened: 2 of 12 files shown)


examples/simple-run/index.ts
```diff
@@ -2,7 +2,7 @@ import * as local from "@pulumi/command/local";
 import * as random from "@pulumi/random";
 import { interpolate } from "@pulumi/pulumi";
 
-const pw = new random.RandomPassword("pw", { length: 10 });
+const pw = new random.RandomPassword("pw", { length: 10, special: false });
```
dixler, line 5: Just to be sure, is this to keep unexpected special shell characters from messing up commands in the tests?

provider/pkg/provider/remote/commandOutputs.go
```diff
@@ -59,35 +59,22 @@ func (c *CommandOutputs) run(ctx p.Context, cmd string) (string, string, error)
...
+	<-stdouterrch
 
-	<-stdoutch
-	<-stderrch
+	if err != nil {
+		return "", "", fmt.Errorf("%w: running %q:\n%s", err, cmd, stdouterrbuf.String())
+	}
 
-	return stdoutbuf.String(), stderrbuf.String(), err
+	return strings.TrimSuffix(stdoutbuf.String(), "\n"), strings.TrimSuffix(stderrbuf.String(), "\n"), nil
```
dixler, line 79: It looks to me that we're stripping newlines now.

dixler (approved):
LGTM. Noted some of the semantic changes I noticed.

examples/simple-run/index.ts
lukehoban: Yep - that’s right - I saw two test failures caused by this while I was making this change - so just fixed the  test to avoid this.

provider/pkg/provider/remote/commandOutputs.go
lukehoban: Yeah - this was to align behavior with local command which does the same. I believe this difference was accidental in the original implementation.

provider/pkg/provider/local/commandOutputs.go
```diff
@@ -46,18 +46,13 @@ func (c *CommandOutputs) run(ctx p.Context, command string) (string, string, err
...
-	}
+	var err error
+	var stdoutbuf, stderrbuf, stdouterrbuf bytes.Buffer
+	r, w := io.Pipe()
 
 	cmd := exec.CommandContext(ctx, args[0], args[1:]...)
-	cmd.Stdout = stdoutw
-	cmd.Stderr = stderrw
+	cmd.Stdout = io.MultiWriter(&stdoutbuf, &stdouterrbuf, w)
+	cmd.Stderr = io.MultiWriter(&stderrbuf, &stdouterrbuf, w)
```
abhinav, line 55: I don't know if this is a concern here, but this may cause a data race in writing to stdouterrbuf because os/exec will spawn separate goroutines to read from the stdout and stderr pipes and write to the provided writers.

os/exec has a special case for when Stdout and Stderr are the same writer, but that won't apply here.

```
	// If Stdout and Stderr are the same writer, and have a type that can
	// be compared with ==, at most one goroutine at a time will call Write.
```

(https://pkg.go.dev/os/exec#Cmd.Stdout)

If that's a concern here, we may want to wrap the bytes.Buffer's Write method with a mutex. Something simple like the following would suffice.

```go
type lockedWriter struct {
  w  io.Writer
  mu sync.Mutex
}

func (w *lockedWriter) Write(bs []byte) (int, error) {
  w.mu.Lock()
  n, err := w.w.Write(bs)
  w.mu.Unlock()
}
```

(This isn't an issue for `io.Pipe` so that `w` shouldn't need this.)

lukehoban: > os/exec has a special case for when Stdout and Stderr are the same writer, but that won't apply here.

Great callout!  Indeed there's an even deeper issue due to this, that we can't guarantee we present a correctly ordered view of the combined stdout+stderr unless we pass them the same writer, and if we do that, we can't break them apart.  So in Go today, it appears there is no way to get both (1) the ability to see the stdout and stderr streams independently and (2) a correctly ordered single stream of stdout+stderr in the order the process emitted them.  Now of course (2) is probably not even 100% well defined at some even lower level, but we can lose arbitrary correctness on this just at the Go level.  After some disbelief around this, I finally found a confirmation that this was a limitation in Go os/exec, but can't find the reference now.

> If that's a concern here, we may want to wrap the bytes.Buffer's Write method with a mutex

Great point on the concurrency concern on the bytes.Buffer, will add the mutex to guard against this.

abhinav (approved):

PR #179, opened 2023-01-17
cmd/gen: Don't take two passes over JSON

abhinav:
MarshalIndent JSON-encodes the input,
and then passes it through `json.Indent` again
to indent it properly.

This isn't necessary.
`json.Encoder` has a `SetIndent` method
that allows opting into indentation in the first pass.

Test Plan:
Verified `make codegen` leaves the schema unchanged.


dixler (approved):
LGTM

PR #180, opened 2023-01-24
Add DialErrorLimit as optional parameter for remote command

adrianriobo:
Previously there was a hard coded value of 10 attempts on dialing the remote host, if 10 attempts returned an error the command execution fails. 

Under some circumstances it could be expected there would be a long period where the host is not responsive but the execution should keep (failing it the case based on defined timeouts). As real use case for that is some setup were a reboot of the machine is required (typically dedicated hosts on AWS as an example will take extra time for the boot) and it makes the remote command fail.

This change just enable to adjust the attempts when getting error dialing, or even just set as unlimited so it is up to the resource timeout to finish the execution. 

Also the new parameter has a default value of 10, so in case the parameter is not set, remote command behaves as it did before


provider/pkg/provider/remote/connection.go
```diff
@@ -52,6 +55,8 @@ func (c *Connection) Annotate(a infer.Annotator) {
 	a.Describe(&c.PrivateKey, "The contents of an SSH key to use for the connection. This takes preference over the password if provided.")
 	a.Describe(&c.PrivateKeyPassword, "The password to use in case the private key is encrypted.")
 	a.Describe(&c.AgentSocketPath, "SSH Agent socket path. Default to environment variable SSH_AUTH_SOCK if present.")
+	a.Describe(&c.DialErrorLimit, "Max allowed errors on trynig to dial the remote host. 0 set count to unlimited. Default value is 10")
```
guineveresaenger, line 58:
```suggestion
	a.Describe(&c.DialErrorLimit, "Max allowed errors on trying to dial the remote host. 0 set count to unlimited. Default value is 10")
```

guineveresaenger (commented):
Thank you for your pull request!
Apologies for the delay in review! 🙏 

Tiny typo fix suggestion.

Holding explicit approval for second review.

provider/pkg/provider/remote/connection.go
```diff
@@ -109,7 +114,7 @@ func (con *Connection) Dial(ctx p.Context, config *ssh.ClientConfig) (*ssh.Clien
 				net.JoinHostPort(*con.Host, fmt.Sprintf("%.0f", *con.Port)),
 				config)
 			if err != nil {
-				if try > 10 {
+				if *con.DialErrorLimit > UnlimitedDialErrors && try > *con.DialErrorLimit {
```
lukehoban, line 117: It looks like perhaps this will panic if `con.DialErrorLimit` is `nil`?  Is this relying on the fact that the `default` is set in the schema to guarantee that a non-nil value will always appear here?  If so, I think the struct property itself can be a non-pointer and non-optional?

provider/cmd/pulumi-resource-command/schema.json
```diff
@@ -471,6 +471,11 @@
                     "description": "SSH Agent socket path. Default to environment variable SSH_AUTH_SOCK if present.",
                     "type": "string"
                 },
+                "dialErrorLimit": {
+                    "default": 10,
+                    "description": "Max allowed errors on trynig to dial the remote host. 0 set count to unlimited. Default value is 10",
```
lukehoban, line 476: Should `-1` be the "infinite" option, so that `0` can be the "don't retry" option?

lukehoban (commented):
Ideally we would have tests here, but I'm not certain how to test this without building a mock SSH server that fails connections :-).  If there is a way to extend the examples/ec2_remote example/test to include usage of this feature, would be great to have at least some coverage in the test suite.

provider/pkg/provider/remote/connection.go
adrianriobo: fixed

adrianriobo:
@lukehoban added a remote command invocation with `dialErrorLimit: 0` on examples/ec2_remote hope this is enough to give a bit coverage, although is far from ideal. If there is a strong requirement on create tests around this I will need a bit more time

provider/cmd/pulumi-resource-command/schema.json
adrianriobo: yeah -1 was my first choice actually and then moved to 0 so totally fine. Already fixed

provider/pkg/provider/remote/connection.go
adrianriobo: Here I left it as it was, yeah DialErrorLimit shoud have value from `default`,  I understood the `optional` exactly because the user has the option to set it, otherwise it will use the default value as documented. For non-pointer I just followed the approach for other values with defaults like `port` or `user`.

iwahbe:
Looks like there is a type error for the ec2 remote example: https://github.com/pulumi/pulumi-command/actions/runs/4176245380/jobs/7232349783

adrianriobo:
@t0yv0  :joy: I am not familiar with`ts` syntax at all, hopefully it was only matter of setting the param name (as it did not match with the property :( ) That is my understanding from logs. :crossed_fingers:

provider/pkg/provider/remote/connection.go
```diff
@@ -109,7 +114,9 @@ func (con *Connection) Dial(ctx p.Context, config *ssh.ClientConfig) (*ssh.Clien
 				net.JoinHostPort(*con.Host, fmt.Sprintf("%.0f", *con.Port)),
 				config)
 			if err != nil {
-				if try > 10 {
+				// on each try we already made a dial
+				dials := try + 1
+				if *con.DialErrorLimit > UnlimitedDialErrors && dials > *con.DialErrorLimit {
```
t0yv0, line 119: Guessing we're counting on *deref not to panic because there's a default value specified.

adrianriobo: yeah well I assume that, just followed same approach as for [`port`](https://github.com/pulumi/pulumi-command/blob/694b12206afdcb529e92d72f30774c3524a5d76f/provider/pkg/provider/remote/connection.go#L109)

adrianriobo:
ok so `0` seems to strict value I will change to `-1` for `unlimited` testing

t0yv0:
That sounds good, I guess 0 made that test not issue any calls at all? Let's try another value and once it's in I'll run acceptance tests again. Thank you!

adrianriobo:
> That sounds good, I guess 0 made that test not issue any calls at all? Let's try another value and once it's in I'll run acceptance tests again. Thank you!

Well `0` means not errors allowed, so it would dial 1 time and if that dial end up on error it finish with error. I guess 0 was too strict...and maybe that is the reason for that random `10` value ? Now I am testing the unlimited option `-1` which is my real case as with unlimited attempts that should fallback to the timeout of the resource I hope....

t0yv0:
Sorry a bit distracted today, looks like we're having to iterate a lot. I'll try to dive deeper tomorrow if it's still not working for some reason, have been doing a very shallow pass here.

adrianriobo:
Ok so I am a bit confused right now, I  tested the branch for `go sdk` and I see the expected behavior:

I used the latest [changes](https://github.com/adrianriobo/qenvs/commit/dfb3fc8d9c392d91c27597a084e888f0ceb10a2e)  and then set [`DialErrorLimit` to `-1`](https://github.com/adrianriobo/qenvs/pull/51/commits/ff7c013730ed22f43eb69a8d65461fbe8d9b0c04)

Finally create some instance with `quay.io/ariobolo/qenvs:dial_limit`, which contains all these,  and command wait as expected, without any error or only one attempt (which seems it is the case for nodejs example?):

```bash
...
{"level":"info","msg":"@ updating....","time":"2023-02-16T11:30:00Z"}
{"level":"info","msg":" +  command:remote:Command ol-rhel-wait creating (0s) ","time":"2023-02-16T11:30:00Z"}
{"level":"info","msg":"@ updating....","time":"2023-02-16T11:30:01Z"}
{"level":"info","msg":" +  aws:lb:Listener crcqe-rhel-ol-r created (1s) ","time":"2023-02-16T11:30:01Z"}
{"level":"info","msg":"@ updating.....................................................................................................................................","time":"2023-02-16T11:32:11Z"}
{"level":"info","msg":" +  command:remote:Command ol-rhel-wait creating (131s) ping","time":"2023-02-16T11:32:11Z"}
{"level":"info","msg":" +  command:remote:Command ol-rhel-wait created (131s) ping","time":"2023-02-16T11:32:11Z"}
...
```

Do you have any thought on what could cause the issue on the `nodejs` example?

t0yv0:
Flake perhaps or an issue with Pulumi tests account. The issue is

```
error: dial tcp 54.185.231.251:22: connect: connection refused
[286](https://github.com/pulumi/pulumi-command/actions/runs/4189324979/jobs/7261541342#step:24:289)
```

I'll try to run again see if it sticks.

t0yv0:
Hm, yes it passed for me locally but keeps failing in CI.

t0yv0:
It passed this time. Seems like a possibly flaky test. Merging! Thank you for your contribution.

PR #188, opened 2023-02-28
Added default value for DialErrorLimit for backwards compatibility on client / plugin versions mismatch

### LINKED ISSUE #187: DialErrorLimit create error on mismatching plugin versions
### What happened?

Recently #180 introduced a new parameter `DialErrorLimit` that parameter has a default value which [prevents from panic](https://github.com/pulumi/pulumi-command/blob/e0d039d1c343e83b779f7a8fbf13726d6ef49ba6/provider/pkg/provider/remote/connection.go#L119) 

The main problem is that default value is set on the client side and the check is done on the plugin side. So in a situation when code is based on previous versions and the plugin is dynamically installed (based on latest version) the plugin will try to access a value which does not exist:

```bash
77    panic: runtime error: invalid memory address or nil pointer dereference
78    [signal SIGSEGV: segmentation violation code=0x1 addr=0x0 pc=0xdfa090]
79    goroutine 34 [running]:
80    [github.com/pulumi/pulumi-command/provider/pkg/provider/remote.(*Connection).Dial.func1(0x0](http://github.com/pulumi/pulumi-command/provider/pkg/provider/remote.%28*Connection%29.Dial.func1%280x0), 0xc00010f068?)
81        /home/runner/work/pulumi-command/pulumi-command/provider/pkg/provider/remote/connection.go:119 +0x1d0
```

### Expected Behavior

When a newer plugin version is installed it should not break the usage with old client code, so new parameters should check if they are not passed from the client and if that is the case the default value should be set at the plugin side.

### Steps to reproduce

Pick code for command provider version `0.7.0` and let pulumi handle the plugin installation (it will pick `0.7.1`) it will create the above error 

### Output of `pulumi about`

N/A

### Additional context

_No response_

### Contributing

adrianriobo:
This PR will fix the issue https://github.com/pulumi/pulumi-command/issues/187 when new version of the plugin (>= v0.7.1) is used with old client classes (<= v0.7.0).  This PR will ensure if no default value is giving to the DialLimitError from the client the default value is set at the plugin side.

Fixes #187

### DIFF
--- a/provider/pkg/provider/remote/connection.go
+++ b/provider/pkg/provider/remote/connection.go
@@ -29,8 +29,11 @@ import (
 
 const (
 	sshAgentSocketEnvVar = "SSH_AUTH_SOCK"
-	DialErrorDefault     = 10
-	DialErrorUnlimited   = -1
+)
+
+var (
+	dialErrorDefault   = 10
+	dialErrorUnlimited = -1
 )
 
 type Connection struct {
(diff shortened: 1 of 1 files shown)


adrianriobo:
@iwahbe :thinking: dunno why those tests keep failing but I tested it out again with go library and the new generated plugin with default value and all went ok, similar situation on previous PR https://github.com/pulumi/pulumi-command/pull/180#issuecomment-1435422527

iwahbe:
@adrianriobo Tests passed after a couple of re-runs. I'm merging now.

Issue #190, opened 2023-03-21
Update Command does not receive previous command stdout or stderr


brucetan-displayr:
### What happened?

The update command could not retrieve the environment variable `$PULUMI_COMMAND_STDOUT`, even though the documentation specifically  mention that `The environment variables PULUMI_COMMAND_STDOUT and PULUMI_COMMAND_STDERR are set to the stdout and stderr properties of the Command resource from previous create or update steps.`

https://www.pulumi.com/registry/packages/command/api-docs/local/command/#update_nodejs

### Expected Behavior

the update command can retrieve environment variables PULUMI_COMMAND_STDOUT and PULUMI_COMMAND_STDERR

### Steps to reproduce

Given the following code
```
const mktemp = new local.Command('mktemp', {
    create: 'mktemp',
    update: 'echo $PULUMI_COMMAND_STDOUT',
    delete: 'rm $PULUMI_COMMAND_STDOUT'
})

```

1. run `pulumi up` to create the resource, at this point, `stdout` would be something like `/tmp/tmp.cPqLweTEcN`
2. add space somewhere to force an update 
3. `stdout` of the resource become empty

### Output of `pulumi about`

CLI
Version      3.57.1
Go Version   go1.20.1
Go Compiler  gc

Plugins
NAME     VERSION
command  0.7.1
nodejs   unknown
random   4.12.0

Host
OS       ubuntu
Version  22.04
Arch     x86_64

This project is written in nodejs: executable='/home/bruce/.nvm/versions/node/v18.14.0/bin/node' version='v18.14.0'

Backend
Name           pulumi.com
URL            https://app.pulumi.com/brucetan-displayr
User           brucetan-displayr
Organizations  brucetan-displayr, displayr

### Contributing

Vote on this issue by adding a 👍 reaction. 
To contribute a fix for this issue, leave a comment (and link to your pull request, if you've opened one already).

brucetan-displayr:
PR that fixes this: https://github.com/pulumi/pulumi-command/pull/191

t0yv0:
Thanks so much for providing the fix! We will release a new version shortly that publishes the fix.

brucetan-displayr:
> Thanks so much for providing the fix! We will release a new version shortly that publishes the fix.

Thanks :D

brucetan-displayr:
hi @t0yv0 , I wonder if there's a timeframe to release the new version? It's been 3 weeks since the PR has been merged.

[...]

PR #191, opened 2023-03-21
Fix stdout and stderr missing in pulumi update command

brucetan-displayr:
This PR fixes this issue: https://github.com/pulumi/pulumi-command/issues/190

In the Update method, the `news` input struct contains Triggers, Create, Update, and Delete fields, while the `olds` output struct contains Stdout, Stderr, Assets, and Archive fields.

The state variable in the Update method only contains `news` input, causing empty stdout and stderr in downstream code https://github.com/pulumi/pulumi-command/blob/master/provider/pkg/provider/local/commandOutputs.go#L71

We can fix this by merging the `news` and `olds` inputs and outputs into a new CommandOutputs struct. The result is a new CommandOutputs struct that contains all the fields from both the CommandInputs and BaseOutputs structs. If any fields are present in both structs, the value from news will take precedence over the value from olds.

### DIFF
--- a/examples/delete-from-stdout/index.ts
+++ b/examples/delete-from-stdout/index.ts
@@ -2,6 +2,7 @@ import { local } from "@pulumi/command";
 
 const mktemp = new local.Command('mktemp', {
     create: 'mktemp',
+    update: 'echo $PULUMI_COMMAND_STDOUT',
     delete: 'rm $PULUMI_COMMAND_STDOUT'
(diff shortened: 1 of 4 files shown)


iwahbe:
@brucetan-displayr Thanks for the PR! This looks good.

t0yv0:
Thank you for your contribution! This looks good. Running tests now.

PR #132, opened 2022-09-14
Allow usage of bastion host for remotecommand and remotefilecopy

nicklasfrahm:
Closes #131.

PR #192, opened 2023-03-22
Allow usage of bastion host for remotecommand and remotefilecopy / Migrated to new provider structure

eBeyond:
I've migrated the work on PR #132  to the new provider structure. (Can't test it currently)


provider/pkg/provider/remote/connection.go
```diff
@@ -107,16 +117,73 @@ func (con *Connection) SShConfig() (*ssh.ClientConfig, error) {
 	return config, nil
 }
 
+func (con *Connection) ProxySShConfig() (*ssh.ClientConfig, error) {
+	config := &ssh.ClientConfig{
+		User:            *con.ProxyUser,
```
lukehoban, line 122: The tests at https://github.com/pulumi/pulumi-command/actions/runs/4527001624/jobs/7972532321#step:25:321 appear to be triggering a panic here.

Looks like we aren't handling the cases where there is no proxy configuration being provided.

provider/cmd/pulumi-resource-command/schema.json
```diff
@@ -497,6 +497,26 @@
...
+                    "type": "integer"
+                },
+                "proxyPrivateKey": {
+                    "description": "The contents of an SSH key to use for the bastion host to setup the connection. This takes preference over the password if provided.",
+                    "type": "string"
+                },
+                "proxyUser": {
+                    "description": "The user that we should use for the bastion host connection.",
+                    "type": "string"
+                },
```
lukehoban, line 519: I think this is probably good the way it is - but a few related questions/thoughts:
1. Should this be nested under a single `proxy` property?
2. There is a lot of overlap with `Connection` itself - should this even be a nested Connection object?
3. There are a couple things on `Connection` that aren't here - `privateKeyPassword`, `dialErrorLimit`.  Are either of these relevant to proxy scenario?

All that said - I think most likely it's best to keep it simple like it is here.

provider/pkg/provider/remote/connection.go
```diff
@@ -107,16 +117,73 @@ func (con *Connection) SShConfig() (*ssh.ClientConfig, error) {
...
 	var client *ssh.Client
+	var proxyClient *ssh.Client
 	var err error
 	var dialErrorLimit = con.getDialErrorLimit()
 	_, _, err = retry.Until(ctx, retry.Acceptor{
 		Accept: func(try int, nextRetryTime time.Duration) (bool, interface{}, error) {
 			client, err = ssh.Dial("tcp",
 				net.JoinHostPort(*con.Host, fmt.Sprintf("%.0f", *con.Port)),
 				config)
+			if *con.ProxyHost != "" {
```
lukehoban, line 156: Does it make sense to only construct the proxyConfig here instead of passing it in?

lukehoban (commented):
One test failing which looks like it neeeds to be addressed.

I think we will need test coverage given this is a non-trivial new extension to the API.  I expect it is possible to set up a test that uses this - either as an optional part of https://github.com/pulumi/pulumi-command/blob/master/examples/ec2_remote/index.ts, or as a separate similar test.

provider/cmd/pulumi-resource-command/schema.json
eBeyond: Currently refactoring it:
To 1: I've refactored it to have the properties below the proxy property
To 2: There is a .Default method in connection wich will be called recursively when using the Connection as a type for the proxy property as such I've created a new type called ProxyConnection (the Connection type has got the proxy field)
To 3: Yes they're relevant, currently in progress to implement them

provider/pkg/provider/remote/connection.go
eBeyond: Refactored as suggested

eBeyond:
Added a test like ec2_remote with proxy ec2_remote_proxy

eBeyond:
Is there any work to do on it?

sy-bee:
Any news on this PR?

eBeyond:
Ping @lukehoban

iwahbe:
@mikhailshilkov This PR looks good to me, but it needs some code cleanup and a rebase to resolve conflicts. I opened https://github.com/pulumi/pulumi-command/pull/269 to get it over the finish line. 

My plan is to leave this PR (and https://github.com/pulumi/pulumi-command/pull/132) open until #269 merges.

My hope is that we can get this out either today or Monday.

iwahbe:
This PR merged as part of https://github.com/pulumi/pulumi-command/pull/269. A big thanks goes to @eBeyond and @nicklasfrahm. They did a lot of the work here.

PR #194, opened 2023-03-23
Add DialTimeout option to remote commands

vboulineau:
By default, in Go, Dial has no timeout, which mean that we may wait up to OS timeout (~2/3 minutes on Linux) during the first attempt while remote system system may be ready earlier.

With `DialErrorLimit`, it also provides better control on the maximum way time (even though the `Retry` parameters, 100ms base + 1.5x backoff are not publicly documented).


vboulineau:
👋 Any update on going forward with this PR?

mikhailshilkov:
@iwahbe Can you help unblock this as a part of your Command investments?

iwahbe:
This functionality merged as part of https://github.com/pulumi/pulumi-command/pull/268.

PR #219, opened 2023-07-14
Re-sync generated sdks

iwahbe:
We haven't regenerated the SDK in some time, and it has drifted with new version of pulumi/pulumi codegen. To see green tests, we need to re-align the checked-in SDK with what pulumi/pulumi currently generates.


sdk/go/command/local/command.go
```diff
@@ -99,10 +100,7 @@ func NewCommand(ctx *pulumi.Context,
 		args = &CommandArgs{}
 	}
 
-	replaceOnChanges := pulumi.ReplaceOnChanges([]string{
-		"triggers[*]",
-	})
```
iwahbe, line 104: This looks like a bug as well. The schema hasn't changed, so we should see the same `replaceOnChanges` behavior here.

PR #229, opened 2023-07-24
Downgrade towards green CI

iwahbe:
CI on `main` is currently failing. It looks like one of our tests is flakey under the new version, and getting worse on later versions of Pulumi. We can downgrade to get back to green.


iwahbe:
> This seems like an odd fix. Downgrading affects the whole provider, whereas we could ignore the flakey test and leave everything else alone?

If the test was recently added, I would agree with you. My concern is that since the update made the test flakey, our users will experience the same flakiness as the test is experiencing.

thomas11 (approved):

iwahbe:
> > > This seems like an odd fix. Downgrading affects the whole provider, whereas we could ignore the flakey test and leave everything else alone?
> > 
> > 
> > If the test was recently added, I would agree with you. My concern is that since the update made the test flakey, our users will experience the same flakiness as the test is experiencing.
> 
> Ah, so the issue isn't just green CI like that title says. That's what through me off. Do we have an issue to investigate this, particularly since there might be user impact?

There hasn't been a release, so there hasn't been user impact. I'll open an issue.

PR #234, opened 2023-07-25
Upgrade pulumi go provider to v0.10.1

### LINKED ISSUE #233: Unable to upgrade past `go-provider-sdk` v0.9.0
### What happened?

Upgrading past v0.9.0 causes `TestEc2RemoteTs` to [fail](https://github.com/pulumi/pulumi-command/actions/runs/5622606375/job/15295799597):

```
  Diagnostics:
    command:remote:Command (hostname):
      error: dial tcp 54.189.207.111:22: connect: connection refused
```

### Expected Behavior

Upgrades without breaking changes don't cause test failures.

### Steps to reproduce

See CI test.

### Output of `pulumi about`

We are on pulumi v3.75.0.

### Additional context

_No response_

### Contributing

iwahbe:
This change is best reviewed on a commit by commit basis. This PR is the result of debugging the `TestEc2RemoteTs` failures in CI on my local machine.

Fixes #233 

The "fix" is in https://github.com/pulumi/pulumi-command/pull/234/commits/0121adfca45bae7c5a2d73ac2a4947c97b38a1ab. I add a new remote command resource: "poll" with unlimited retries to access the deployed server.[^1] All other remote command resources (with either 1 or 10 connection attempts) are given a "dependsOn" relationship to "poll". This ensures that we don't try to connect to the server until it is fully deployed.

[^1]: This resource has a 10m timeout. This should cause genuine deployment failures to fail instead of hanging. Since most server test runs take 3-5 minutes and that creates two servers (in sequence), this should prevent slow deployments from failing tests.


examples/ec2_remote/index.ts
```diff
@@ -42,31 +45,38 @@ const connection: types.input.remote.ConnectionArgs = {
 };
 
 const connectionNoDialRetry: types.input.remote.ConnectionArgs = {
-    host: server.publicIp,
-    user: "ec2-user",
-    privateKey: privateKey,
-    dialErrorLimit: -1,
```
lukehoban, line 48: Why are all the changes here necessary?  Shouldn’t the `-1` here have accomplished retrying the connection already?

examples/ec2_remote/index.ts
```diff
@@ -42,31 +45,38 @@ const connection: types.input.remote.ConnectionArgs = {
...
 };
 
+// We poll the server until it responds.
+//
+// Because other commands depend on this command, other commands are guaranteed
+// to hit an already booted server.
+const poll = new remote.Command("poll", {
+  connection: { ...connection, dialErrorLimit: -1 },
+    create: "echo 'Connection established'",
+}, { customTimeouts: { create: "10m" } })
```
lukehoban, line 59: This doesn’t feel like it should be something we require users to do. But it’s also not clear why it helps.  It seems to just do what the code did for the other commands prior to the changes here?

examples/ec2_remote/index.ts
iwahbe: I changed the definition of `connectionNoDialRetry` because it was called `connectionNoDialRetry` and it had the effect of granting infinite dial retries. This change isn't necessary to get the test to pass, it just makes more sense to read. This test is also an example.

examples/ec2_remote/index.ts
lukehoban: > It is required because aws.ec2.Instance returns before the resource is fully up and running. If we want to make this unnecessary the fix will need to be in aws and ultimately in pulumi-aws's upstream.

But prior to these changes, all resources used `dialErrorLimit: -1,` already, right?  So should have been waiting for it to be ready even with no changes?  

It seems the changes to the example moved from (a) all resources use `dialErrorLimit: -1,` to (b) one resources uses `dialErrorLimit: -1,`, and the rest depend on that resource.  I can't tell why that helps.  Why didn't (a) also work for the same reasons (b) does?

lukehoban: My concern is - are we saying that to correctly use these resources you have to write code like this?  I hope not - it seems like it is supposed to be the job of `remote.Command` to ensure it can retry for a sufficiently long time to ensure this works in just the "obvious" way?

lukehoban (approved):
Happy to merge, as this unblocks releasing some important fixes.  But still unsure whether we are saying it is fundamentally more complex to use `remote.Command` now, or whether this example is just testing of both `dialErrorLimit: -1,` and `dialErrorLimit: 1,` in a single test.

examples/ec2_remote/index.ts
iwahbe: > But prior to these changes, all resources used dialErrorLimit: -1, already, right? So should have been waiting for it to be ready even with no changes?

No. Previously `"hostname"`, `"remotePrivateIP"`, `"size"` and `"checkSize"` (using logical names) all used `connection` to connect. `connection` is defined as:

https://github.com/pulumi/pulumi-command/blob/f0ba2f06be7304bff459cad77bc626d0596529a3/examples/ec2_remote/index.ts#L38-L42

`connection` doesn't specify `dialErrorLimit`, so it defaults to `10`. That's why it doesn't wait forever.

`hostname` was the resource that generally errored. The resources defined with `{ dialErrorLimit: -1 }` did not error.

> My concern is - are we saying that to correctly use these resources you have to write code like this? I hope not - it seems like it is supposed to be the job of remote.Command to ensure it can retry for a sufficiently long time to ensure this works in just the "obvious" way?

It was always necessary to write code like this to remove the race condition. In the past we tended to see the race condition work out in our favor, but it was still there. We would see occasional test failures that were fixed with retry.

We can extend the default retry limit, but if your service takes an hour to become responsive, I don't think this should work by default. If you misconfigure your network so the connection will always be refused, it shouldn't take an hour of retries to fail.

Wait for the server to provision and fail fast are both important and in tension here.

iwahbe:
> Happy to merge, as this unblocks releasing some important fixes. But still unsure whether we are saying it is fundamentally more complex to use `remote.Command` now, or whether this example is just testing of both `dialErrorLimit: -1,` and `dialErrorLimit: 1,` in a single test.

The latter. It is exactly as complex to use `remote.Command` now as it was in the past.

Issue #198, opened 2023-05-25
`run` does not return stdout or stderr


Frassle:
### What happened?

See https://github.com/pulumi/pulumi/discussions/13013#discussioncomment-5996365 for initial user report on this.
I've tried repro'ing myself and can confirm that stdout and stderr are returned empty even for simple commands such as `echo hi`.

### Expected Behavior

stdout and stderr to contain the output of the command ran.

### Steps to reproduce

```
import * as command from "@pulumi/command";

// Render the Shell Script
const renderShell1 = command.local.run({
    command: "echo hello",
    environment: {
        //IMAGE_URL: imageurl,
        IMAGE_URL: "test123",
    },
  });
export const stdo = renderShell1.then(v => v.stdout);
export const stde = renderShell1.then(v => v.stderr);
```

### Output of `pulumi about`

`@pulumi/command  0.7.2`

### Contributing

Vote on this issue by adding a 👍 reaction. 
To contribute a fix for this issue, leave a comment (and link to your pull request, if you've opened one already).

aq17:
@dixler would you be able to take a look at this as a command provider expert 😬

PR #246, opened 2023-08-03
Fix stdout and stderr on run command

aq17:
Mimics https://github.com/pulumi/pulumi-command/pull/191/files but on create, so `stdout` + `stderr` is emitted on the initial create as well as subsequent updates

Fixes https://github.com/pulumi/pulumi-command/issues/198

### DIFF
--- a/provider/pkg/provider/local/commandController.go
+++ b/provider/pkg/provider/local/commandController.go
@@ -31,7 +31,8 @@ var _ = (infer.ExplicitDependencies[CommandInputs, CommandOutputs])((*Command)(n
 
 // This is the Create method. This will be run on every Command resource creation.
 func (c *Command) Create(ctx p.Context, name string, input CommandInputs, preview bool) (string, CommandOutputs, error) {
-	state := CommandOutputs{CommandInputs: input}
+	outputs := BaseOutputs{Stdout: *input.Stdin, Stderr: *input.Stdin}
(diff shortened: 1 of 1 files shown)


provider/pkg/provider/local/commandController.go
```diff
@@ -31,7 +31,8 @@ var _ = (infer.ExplicitDependencies[CommandInputs, CommandOutputs])((*Command)(n
 
 // This is the Create method. This will be run on every Command resource creation.
 func (c *Command) Create(ctx p.Context, name string, input CommandInputs, preview bool) (string, CommandOutputs, error) {
-	state := CommandOutputs{CommandInputs: input}
+	outputs := BaseOutputs{Stdout: *input.Stdin, Stderr: *input.Stdin}
```
iwahbe, line 34: I don't understand why we are setting `stderr` to `*input.stdin`. `stderr` should not be set until a command is actually run.

iwahbe (commented):
Can you tag the bug you are trying to fix? We will need a test that confirms that this works as expected.

I don't understand why this fix changes the behavior in a meaningful way, especially since `state.Stdout` and `state.Stderr` are set later on line `49`.

smulikHakipod:
@iwahbe I think it fixes this https://github.com/pulumi/pulumi-command/issues/198 issue

iwahbe:
Thanks @smulikHakipod.

Issue #166, opened 2022-12-22
regression on command 0.7.0: stdout can not be loaded as json using json.loads()


wuxxin:
### What happened?

- command plugin updated from 0.6.0 to 0.7.0
- pulumi up aborts with error message that it can not parse the supposed json string at char 0 line 0
- downgrading command to 0.6.0 and modifying pulumi stack with the resourceid and version of the old command plugin fixes the issue

### Steps to reproduce

the corresponding code snippet is as following:

```python
vault_ca = command.local.Command(
    "{}_vault_ca".format(name),
    create="./__vault_pipe.sh --yes",
    stdin=json.dumps(vault_config),
    opts=pulumi.ResourceOptions(
        parent=self,
        additional_secret_outputs=["stdout"],
        protect=True,
        # XXX make pulumi find legacy name for ca_factory_fault_ca
        aliases=[Alias(name="vault_ca")] if name == "ca_factory" else [],
    ),
)
ca_secrets = vault_ca.stdout.apply(lambda x: json.loads(x))
```

pulumi fails on ca_secret = ... while trying lambda x: json.loads(x)

it is probably reproducable by calling any external command, which should return a json formatted output to be parse by pulumi

### Expected Behavior

same functionality as 0.6.0

### Actual Behavior

pulumi bails out while trying to read stdout of the command as json

### Output of `pulumi about`

CLI
Version      3.46.1
Go Version   go1.19.3
Go Compiler  gc

Host
OS       arch
Version  22.0.0
Arch     x86_64

Backend
Name           power
URL            file:///home/wuxxin/code/athome/state/pulumi
User           wuxxin
Organizations

### Contributing

Vote on this issue by adding a 👍 reaction. 
To contribute a fix for this issue, leave a comment (and link to your pull request, if you've opened one already).

mikhailshilkov:
@dixler Could you please take a look?

AMKamel:
I have the same issue

dixler:
Hi. Thanks @wuxxin for filing the issue with the repro code (it helped me write a test case) and @AMKamel for the context it really helped narrow down what was going on. When moving from 0.6.0 to 0.7.0, fields being computed were dropped and that explains why it was misbehaving in preview.

This is an issue in the pulumi-go-provider library. I have opened https://github.com/pulumi/pulumi-go-provider/issues/75 to track this.

jkodroff:
[...]

PR #251, opened 2023-08-23
Bump go-provider-sdk

iwahbe:
Fixes https://github.com/pulumi/pulumi-command/issues/166

I have added tests that confirm that #166 works as expected.

I have have removed the explicit dependency wiring from `local.Command`, since it matches the default behavior.

I have updated the go version we build against to 1.21.x, since 1.19 is no longer supported.

I have fixed `make lint` so it runs, and made two fixes that `gosimple` suggested.


provider/tests/provider_test.go
```diff
@@ -0,0 +1,112 @@
...
+		"create": resource.PropertyValue{V: "echo hello, $NAME!"},
+		"stderr": c(resource.PropertyValue{V: ""}),
+		"stdout": c(resource.PropertyValue{V: ""}),
+	},
+		create(true /* preview */, unknown).Properties)
+
+	fullProperties := resource.PropertyMap{
+		"create": resource.PropertyValue{V: "echo hello, $NAME!"},
+		"stderr": resource.PropertyValue{V: ""},
+		"stdout": resource.PropertyValue{V: "hello, world!"},
```
rquitales, line 70: A bit unfamiliar with this method of testing. Could you explain why `hello, world!` is stdout here? Perhaps some comments about what each major step is doing might be beneficial.

iwahbe: `"stdout"` is `"hello, world!"` because the command provider runs `echo hello, $NAME!` with `NAME=world` set in the enclosing environment.

iwahbe: There aren't really major steps here. Instead it's a series of logically (but not semantically) tests. I'll refactor to make that clearer.

rquitales (approved):

PR #258, opened 2023-09-02
Add a dispatch workflow to regenerate SDKs

iwahbe:
Pulumi no longer ships SDK generation as a single binary, and it is consequently harder to keep your local system in sync. To simplify SDK generation, we should allow regenerating SDKs in CI. Ideally, we can automate SDK regeneration after a new pulumi version comes out.


t0yv0:
Well this is quite nice, let's have it.

QQ. Why is this not in ci-mgmt would it be trivial to drop it in there?

I'm going to link in https://github.com/pulumi/ci-mgmt/issues/535 here that imagines some more improvements around sdk generation though that one clearly has more scope.

t0yv0 (approved):

iwahbe:
> Well this is quite nice, let's have it.
> 
> QQ. Why is this not in ci-mgmt would it be trivial to drop it in there?
> 
> I'm going to link in [pulumi/ci-mgmt#535](https://github.com/pulumi/ci-mgmt/issues/535) here that imagines some more improvements around sdk generation though that one clearly has more scope.

I wrote this workflow because I couldn't figure out how to reproduce a clean build locally, and I needed to upgrade pulumi-command (which uniquely floats codegen with Pulumi's latest). This repo doesn't use ci-mgmt, so I wouldn't get any benefit here. If there is interest, I would be happy to upstream this into ci-mgmt.

t0yv0:
Yes I think upstreaming to ci-mgmt is worth the 30min or so it would take.

PR #262, opened 2023-09-08
Pin pulumi CLI

t0yv0:
Borrowing targets from pulumi-azure-native here to make sure code generation is pinned.


t0yv0:
Guessing weekly-pulumi-update needs a small change too.

t0yv0:
Oh this is not hermetic.

```
t0yv0@Antons-MacBook-Pro> make python_sdk
rm -rf sdk/python
.pulumi/bin/pulumi package gen-sdk --language python provider/cmd/pulumi-resource-command/schema.json
warning: using pulumi-language-python from $PATH at /Users/t0yv0/.nix-profile/bin/pulumi-language-python
error: method GeneratePackage not implemented
make: *** [python_sdk] Error 255

```

t0yv0:
I've added a fix to make sure language plugins are hermetic and governed by .pulumi.version.

t0yv0:
This is made hermetic now with follow-up commit. Thanks.

mjeffryes (approved):
nice

t0yv0:
Filed https://github.com/pulumi/pulumi-command/issues/265 for later.

Makefile
```diff
@@ -40,34 +43,34 @@ test_provider::
 	cd provider/pkg && go test -short -v -count=1 -cover -timeout 2h -parallel ${TESTPARALLELISM} ./...
 
 dotnet_sdk:: DOTNET_VERSION := $(shell pulumictl get version --language dotnet)
-dotnet_sdk::
+dotnet_sdk::	.pulumi/bin/pulumi
 	rm -rf sdk/dotnet
-	pulumi package gen-sdk --language dotnet $(SCHEMA_FILE)
+	.pulumi/bin/pulumi package gen-sdk --language dotnet $(SCHEMA_FILE)
```
iwahbe, line 48: If we are adding this to `PATH`, we shouldn't need to prefix `pulumi`.

t0yv0: See https://github.com/pulumi/pulumi-command/pull/266

PR #267, opened 2023-09-12
Simplify `run`

### LINKED ISSUE #198: `run` does not return stdout or stderr

iwahbe:
By moving `run` from a method on `*CommandOutputs` to it's own function, we make it obvious what it effects. This has the advantage of making it operate seamlessly on both `CommandOutputs` and `RunOutputs`.

Fixes #198


provider/pkg/provider/local/commandController.go
```diff
@@ -81,6 +80,5 @@ func (c *Command) Delete(ctx p.Context, id string, props CommandOutputs) error {
 	if props.Delete == nil {
 		return nil
 	}
-	_, _, err := props.run(ctx, *props.Delete)
-	return err
+	return run(ctx, *props.Delete, &props.BaseInputs, &props.BaseOutputs)
```
mjeffryes, line 83: this is the bit that specifically fixes the linked issue right?

mjeffryes (approved):
I don't see any problems, and the additional test is nice, but I have to confess that I'm not entirely clear on the goal of the refactor? It looks to be pretty equivalent in complexity before and after. I guess you do get to enforce the invariant that there's a non-empty destination for stderr and stdout in a single place, but on average, functions that mutate their inputs tend to be more complicated to reason about not less. 

I'm ok with shipping this, but I think my preference would be to fix the bug without the refactor and enforce the desired behavior (of emitting stdout/stderr) at a higher level.

provider/pkg/provider/local/commandController.go
iwahbe: No, the fix is provider/pkg/provider/local/runController.go:L31.

iwahbe:
> I don't see any problems, and the additional test is nice, but I have to confess that I'm not entirely clear on the goal of the refactor? It looks to be pretty equivalent in complexity before and after. I guess you do get to enforce the invariant that there's a non-empty destination for stderr and stdout in a single place, but on average, functions that mutate their inputs tend to be more complicated to reason about not less.
> 
> I'm ok with shipping this, but I think my preference would be to fix the bug without the refactor and enforce the desired behavior (of emitting stdout/stderr) at a higher level.

The refactor/fix is necessary because `run` was a method on `*CommandOutputs`, but the `Run` invoke does't use `*CommandOutputs`. Previously we constructed a partial version of `*CommandOutputs` so we could call `run` on it, then extract the output (which we did incorrectly).

PR #268, opened 2023-09-12
Timeout `ssh.Dial`

### LINKED ISSUE #199: Remote Command should fail if connectivity to remote host can not be established
### What happened?

Trying to use Remote Command, connecting to a dummy ip like "8.8.8.8" with random credentials, do not fail, however keeps running for a long while.

Would expect it to fail with some error cannot connect to remote host, if it can not connect to the remote machine.

Would it be possible to add some default timeout?

I guess one option of handling it on our end would be to use [Custom Timeouts](https://www.pulumi.com/docs/concepts/options/customtimeouts/) , but that would not be optimal if it is actually running some kind of long running job.

### Expected Behavior

After max 60 seconds throw an error that it cannot connect to the remote host.

### Steps to reproduce

```
# Run a script to update packages on the remote machine.
update_os_cmd = command.remote.Command("updatePythonCmd",
    connection=command.remote.ConnectionArgs(
        host="8.8.8.8",
        port=22,
        user="pirates",
        password="johnnydepp",
    ),
    create="""(sudo apt update -y);
(sudo apt upgrade -y);
""")
```

Output - and counting..:
command:remote:Command  updatePythonCmd   updating (716s)...  [diff: ~connection]

### Output of `pulumi about`

CLI          
Version      3.68.0
Go Version   go1.20.4
Go Compiler  gc

Plugins
NAME          VERSION
azure-native  1.102.0
command       0.7.2
python        unknown

Host     
OS       ubuntu
Version  22.04
Arch     x86_64

This project is written in python: executable='/home/linuxbrew/.linuxbrew/bin/python3' version='3.11.2
'

Backend        
Name           pulumi.com
URL            https://app.pulumi.com/x

Dependencies:
NAME                 VERSION
pip                  23.1.2
pulumi-azure-native  1.102.0
pulumi-command       0.7.2
setuptools           67.7.2
wheel                0.40.0

Pulumi locates its logs in /tmp by default

### Additional context

_No response_

### Contributing

iwahbe:
Fixes #199 

By default, `golang.org/x/crypto/ssh.Dial` does not timeout. This sets the default timeout to 15 seconds per attempt.

We expose `perDialTimeout` to the user, allowing them the ability to set the per-dial timeout. Setting `perDialTimeout` to 0 implies no timeout.


provider/tests/go.mod
```diff
@@ -9,7 +9,7 @@ require (
 	github.com/pulumi/pulumi-command/provider v0.0.0-00010101000000-000000000000
 	github.com/pulumi/pulumi-go-provider v0.12.1-0.20230911215630-ea49c5f9e9b1
 	github.com/pulumi/pulumi-go-provider/integration v0.10.0
-	github.com/pulumi/pulumi/sdk/v3 v3.79.0
+	github.com/pulumi/pulumi/sdk/v3 v3.81.0
```
t0yv0, line 12: Updates riding along, always best as separate PRs but I understand the expediency.

provider/pkg/provider/remote/connection.go
```diff
@@ -59,14 +60,17 @@ func (c *Connection) Annotate(a infer.Annotator) {
...
 	a.SetDefault(&c.DialErrorLimit, dialErrorDefault)
+	a.Describe(&c.PerDialTimeout, "Max number of seconds for each dial attempt. 0 implies no maximum. Default value is 15 seconds.")
+	a.SetDefault(&c.PerDialTimeout, 15)
 }
 
 func (con *Connection) SShConfig() (*ssh.ClientConfig, error) {
 	config := &ssh.ClientConfig{
 		User:            *con.User,
 		HostKeyCallback: ssh.InsecureIgnoreHostKey(),
+		Timeout:         time.Second * time.Duration(con.PerDialTimeout),
```
t0yv0, line 73: Nice, time.Second* makes sure the unit is right.

t0yv0 (approved):

t0yv0:
LGTM

PR #269, opened 2023-09-13
Allow Proxy/Bastion host

iwahbe:
Rebase of #192.


iwahbe:
Hey @eBeyond. I'd appreciate if you could take a look and make sure I didn't miss anything while rebasing https://github.com/pulumi/pulumi-command/pull/192.

I made sure we were running the test you included (thanks!) and did some code cleanup.

provider/pkg/provider/remote/connection.go
```diff
@@ -122,8 +128,51 @@ func (con *Connection) Dial(ctx p.Context, config *ssh.ClientConfig) (*ssh.Clien
...
+						return true, nil, err
+					}
+					return false, nil, nil
+				}
+
+				endpoint := net.JoinHostPort(*con.Host, fmt.Sprintf("%d", int(*con.Port)))
+				netConn, err := proxyClient.Dial("tcp", endpoint)
+				if err != nil {
+					// on each try we already made a dial
+					dials := try + 1
```
lukehoban, line 154: Is it intentional that we aren't applying the same logging and timeout adjustments on these paths?  I don't quite follow why those are necessary on the other branch - but seems we may want to be consistent between the two?

examples/ec2_remote_proxy/index.ts
```diff
@@ -0,0 +1,96 @@
...
+        name: "name",
+        values: ["amzn-ami-hvm-*-x86_64-gp2"],
+    }],
+});
+
+const server = new aws.ec2.Instance("server", {
+    instanceType: size,
+    ami: ami.id,
+    keyName: keyName,
+    vpcSecurityGroupIds: [secgrp.id],
```
lukehoban, line 38: Maybe not strictly necessary - but you could use a separate security group that restricts ingress to the proxy servers security group, so that access is not possible directly and must go through the proxy to really prove that this is working (and not just skipping the proxy).

examples/examples_nodejs_test.go
```diff
@@ -237,6 +237,82 @@ func TestEc2RemoteTs(t *testing.T) {
...
+
+					_, cOk := ciphertext.(string)
+					return cOk
+				}
+
+				assertEncryptedValue := func(m map[string]interface{}, key string) {
+					assert.Truef(t, isEncrypted(m[key]), "%s value should be encrypted", key)
+				}
+				assertEncryptedValue(stack.Outputs, "connectionSecret")
+				assertEncryptedValue(stack.Outputs, "secretEnv")
```
lukehoban, line 309: Not necessarily a problem - but we probably don't need to duplicate this part of the testing here - since it's covered in the other tests.

lukehoban (approved):

provider/pkg/provider/remote/connection.go
iwahbe: We should. I'll separate the retry logic out into another function, then apply it as appropriate to all connections.

examples/ec2_remote_proxy/index.ts
iwahbe: I like that idea. Especially since its the "examples" folder, and that is a much more compelling example.

provider/go.mod
```diff
@@ -6,7 +6,7 @@ require (
 	github.com/blang/semver v3.5.1+incompatible
 	github.com/gobwas/glob v0.2.3
 	github.com/pkg/sftp v1.13.4
-	github.com/pulumi/pulumi-go-provider v0.12.1
+	github.com/pulumi/pulumi-go-provider v0.12.2-0.20230914002200-b7679cf2e409
```
t0yv0, line 9: New version coming soon or we need to depend on pre-release here?

provider/pkg/provider/remote/connection.go
```diff
@@ -114,45 +121,97 @@ func (con *Connection) SShConfig() (*ssh.ClientConfig, error) {
...
+
+// Dial a ssh client connection from a ssh client configuration, retrying as necessary.
+func (con *Connection) Dial(ctx p.Context) (*ssh.Client, error) {
+	config, err := con.SShConfig()
 	if err != nil {
 		return nil, err
 	}
-	return client, nil
+
+	endpoint := net.JoinHostPort(*con.Host, fmt.Sprintf("%d", int(*con.Port)))
```
t0yv0, line 169: No panic in `*con.Port` iff con.Port=nil?

provider/pkg/provider/remote/connection.go
```diff
@@ -46,7 +51,7 @@ type Connection struct {
 	PrivateKeyPassword *string  `pulumi:"privateKeyPassword,optional"`
 	AgentSocketPath    *string  `pulumi:"agentSocketPath,optional"`
 	DialErrorLimit     *int     `pulumi:"dialErrorLimit,optional"`
-	PerDialTimeout     int      `pulumi:"perDialTimeout,optional"`
+	PerDialTimeout     *int     `pulumi:"perDialTimeout,optional"`
```
t0yv0, line 54: This is necessary?

provider/pkg/provider/remote/connection.go
```diff
@@ -61,16 +66,17 @@ func (c *Connection) Annotate(a infer.Annotator) {
...
 	a.SetDefault(&c.PerDialTimeout, 15)
 }
 
-func (con *Connection) SShConfig() (*ssh.ClientConfig, error) {
+func (con *connectionBase) SShConfig() (*ssh.ClientConfig, error) {
 	config := &ssh.ClientConfig{
 		User:            *con.User,
 		HostKeyCallback: ssh.InsecureIgnoreHostKey(),
-		Timeout:         time.Second * time.Duration(con.PerDialTimeout),
+		Timeout:         time.Second * time.Duration(*con.PerDialTimeout),
```
t0yv0, line 79: con.PerDialTimeout never nil presumably?

provider/pkg/provider/remote/connection.go
```diff
@@ -114,48 +120,95 @@ func (con *Connection) SShConfig() (*ssh.ClientConfig, error) {
...
+	var userError error
+	_, data, err := retry.Until(ctx, retry.Acceptor{
+		Accept: func(try int, _ time.Duration) (bool, any, error) {
+			var result T
+			result, userError = f()
+			if userError == nil {
+				return true, result, nil
+			}
+			dials := try + 1
+			if reachedDialingErrorLimit(dials, maxAttemptes) {
```
t0yv0, line 133: Nit: maxAttempts

provider/pkg/provider/remote/connection.go
```diff
@@ -114,48 +120,95 @@ func (con *Connection) SShConfig() (*ssh.ClientConfig, error) {
...
+				return true, nil, userError
```
t0yv0, line 134: You might want to pack some information like "After N attempts" into the userError.

provider/pkg/provider/remote/connection.go
```diff
@@ -114,48 +120,95 @@ func (con *Connection) SShConfig() (*ssh.ClientConfig, error) {
...
+		return proxyClient.Dial("tcp", endpoint)
+	})
+	if err != nil {
+		return nil, fmt.Errorf("proxy: %w", err)
+	}
+
+	// We initiate a SSH connection over the bridge we just established.
+	var channel <-chan ssh.NewChannel
+	var req <-chan *ssh.Request
+	proxyConn, err := dialWithRetry(ctx, "Dial", tries, func() (ssh.Conn, error) {
```
t0yv0, line 198: I love that dialWithRetry is labelled with the operation.

t0yv0 (approved):
🚢

provider/pkg/provider/remote/connection.go
iwahbe: It's not necessary, but it is more consistent with the style of the provider and the other optional values near it.

provider/pkg/provider/remote/connection.go
iwahbe: Nope: https://github.com/pulumi/pulumi-command/blob/be77edab46000128798caaf5b67a9673558e75e4/provider/pkg/provider/remote/connection.go#L72

If the user didn't supply it, we would have added the default value.

Issue #256, opened 2023-09-01
Regression in pulumi command makes command.output secret failing examples


t0yv0:
### What happened?

Copying from: https://github.com/pulumi/examples/issues/1491

TestAccAwsPyEc2Provisioners test is failing in CI on master. 

I've looked into this and here's the problem.  There's two versions of this program, one in TS and one in Python.

With recent changes to pulumi-command, if you are on 0.8.2, the following program:

```typescript
// Execute a basic command on our server.
const catConfig = new command.remote.Command("cat-config", {
    triggers: [changeToken],
    connection,
    create: "cat myapp.conf",
}, { dependsOn: cpConfig });

export const publicIp = server.publicIp;
export const publicHostName = server.publicDns;
export const catConfigStdout = catConfig.stdout;
```

now marks catConfig.Stdout as a secret output whereas it didn't before.

Now, Node test pins Pulumi-command 0.0.3 or some such, while Python test floats to the latest reference. Therefore Node test passes but Python test fails. 

The specific failure is a panic when reading the secret value in ProgramTest Go code. There's an assert that tries to convert catConfigStdout to a string, but it is now a map in Go land.  AFAIK Go ProgramTest cannot yet peek inside secret values returned from the stack.
 

### Expected Behavior

Passing test

### Steps to reproduce

N/A

### Output of `pulumi about`

N/A

### Additional context

We had some discussion and dug up this change https://github.com/pulumi/pulumi-go-provider/pull/105/files#diff-4bd4b3f761d7688f63a5a8e766057562d08e3df2f51cfa88e604cb4e353fd042R318 as possible root cause.

**Need to decide** if this semantic is reasonable for Command and we want to keep it or revert back. If keeping this semantic, we need to adjust the test here to not fail on receiving a secret.

I think the examples issue is fixed by pinning older Command version which is unsustainable so if we decide to keep this behavior we should update the test to unpin.

### Contributing

Vote on this issue by adding a 👍 reaction. 
To contribute a fix for this issue, leave a comment (and link to your pull request, if you've opened one already).

mikhailshilkov:
It sounds like we need to decide whether this is desired behaviour or not. I'll mark this issue as P1 as a regression for now - until we close it one way or the other.

@iwahbe Can you summarize the change that caused the issue and the reasoning behind making `output` a secret?

t0yv0:
[...]

PR #274, opened 2023-09-21
Use default implementation for input-output dependencies

iwahbe:
WireDependencies has quite a but of subtlety at this point. It needs to account for flowing computedness and secretness between inputs and outputs, each of which can be handled either together or separately. When this provider was implemented, the default implementation didn't work as expected, and didn't cover cases such as pass through inputs.

The pulumi-go-provider now does a much better job by default, so we can rely on that implementation. We expect the following semantics:
- If any input is computed, then all outputs are computed.
  - This does not apply to an output `k` if: - There is an input also called `k` and `k` is not computed. - `input[k] == output[k]`.
- If any input is secret, then the matching output of the same name is also secret.

This should fix (not closing) https://github.com/pulumi/pulumi-command/issues/256.
I'll close https://github.com/pulumi/pulumi-command/issues/256 once I merge the examples revert.


t0yv0:
So removing custom configuration to rely on the defaults is looking good! 

I hear this is going to fix https://github.com/pulumi/pulumi-command/pull/274

I would love to see a test added that ensures 274 does not regress.

Frassle:
> If any input is secret, then the matching output of the same name is also secret.

N.B. The engine does this for you anyway.

iwahbe:
> > If any input is secret, then the matching output of the same name is also secret.
> 
> N.B. The engine does this for you anyway.

I thought that the engine only did that if the provider sent `AcceptSecrets: false`. @Frassle Does the engine gate any behavior (besides sending secrets) on `AcceptSecrets: true`?

t0yv0 (approved):
🚢

t0yv0:
> If any input is secret, then the matching output of the same name is also secret.

Privately I think it's a misfeature that caters to Terraform design where every input has a matching output; both in the engine and any provider SDKs we built. I'd prefer dynamic tracing of secrets through black box transforms like TF providers, but for our own providers like command/docker make no black boxes but instead fully accurate secret tracing via Output and Apply.  

But that's a conversation for another day; whatever we have needs to be documented, and whatever we want to have instead needs to undergo change management from status quo which is difficult.

t0yv0:
I'll file the dynamic secrets idea.

Frassle:
> I thought that the engine only did that if the provider sent

No it always does it, it's supposed to help catch providers that didn't correctly flow secrets. Does mean a provider can't unsecret an output "x" if the input "x" is secret.

> Does the engine gate any behavior (besides sending secrets) on AcceptSecrets: true?

No, I think AcceptSecrets just changes if secret objects get sent on the wire protocol.

Frassle:
> I'd prefer dynamic tracing of secrets through black box transforms like TF providers (though Ian notes it has security downsides), but for our own providers like command/docker make no black boxes but instead fully accurate secret tracing via Output and Apply.

Yes I agree, but pragmatism was that we do this in the engine for now.

t0yv0:
https://github.com/pulumi/pulumi/issues/14020 dynamic secrets idea, obviously not actionable for now but to reference in discussion like this.

PR #295, opened 2023-10-31
Update pulumi/pulumi to v3.91.1

mjeffryes:
Doesn't look like there's actually any impact on the python SDK for command, but may as well bump it so we can easily verify it's ok.


iwahbe:
There is. See https://github.com/pulumi/pulumi-command/pull/296.

mjeffryes:
hmm... strange why I did not get the same changes in code gen as you did in #296. Will close in favor of your PR though

iwahbe:
@mjeffryes Did you change .pulumi.version?

mjeffryes:
ah, no I didn't. It's getting a little tricky to keep track of which repos have adopted that convention.

PR #392, opened 2024-04-11
Tie codegen version to `github.com/pulumi/pulumi/pkg/v3` version

iwahbe:
This PR does 3 things:
- Upgrade the version of codegen used to v3.112.0. (first commit)
- Tie the codegen version used to the version of `github.com/pulumi/pulumi/pkg/v3` used. This allows codegen updates to be handled by pulumi-go-provider updates (which can already change the schema, since they have generating machinery)
- Break `make build_%` into separate targets for generate (`make sdk/%`) and build (`make build_%`). This makes it easier for external contributors not set up in python, dotnet, typescript and java to generate SDKs. The all-encompassing target for generation is now `make codegen`

A clean `make codegen` on slow wifi takes ~40 seconds, 36 of which are downloading the pulumi binary. The second run takes 3 seconds, providing contributors a fast hot loop for codegen.


Makefile
```diff
@@ -146,9 +158,19 @@ install_nodejs_sdk::
...
 	cd examples && go test -v -tags=all -timeout 2h
 
-# --------- File-based targets --------- #
-
-.pulumi/bin/pulumi: PULUMI_VERSION := $(shell cat .pulumi.version)
-.pulumi/bin/pulumi: HOME := $(WORKING_DIR)
-.pulumi/bin/pulumi: .pulumi.version
-	curl -fsSL https://get.pulumi.com | sh -s -- --version "$(PULUMI_VERSION)"
+# Keep the version of the pulumi binary used for code generation in sync with the version
+# of the dependency used by github.com/pulumi/pulumi-pulumiservice/provider
```
thomas11, line 162: should that be github.com/pulumi/pulumi-command/provider ?

thomas11 (approved):
very nice!

PR #395, opened 2024-04-12
Don't swallow Setenv error and add guidance to docs

### LINKED ISSUE #48: Remote command `environment` field does not actually set the environment.
## Hello!

- To contribute a fix for this issue, leave a comment (and link to your pull request, if you've opened one already)

## Issue details

The `environment` field does not actually set the environment. This is using pulumi-command version 0.0.3.

### Steps to reproduce

1. Try to run a remote command while setting the environment.
```python3
connection = command.remote.ConnectionArgs(
    host=instance.public_ip,
    user="ubuntu",
    private_key=config.require_secret("ssh_private_key"),
)
command.remote.Command(
    "testvar",
    connection=connection,
    create='echo "${TESTVAR}" > testvar',
    environment={
        "TESTVAR": "asdf1234",
    },
)
```

Expected: The environment variable should be set and accessible when running the command. The contents of the `testvar` file should contain `asdf1234\n`.
Actual: The environment variable seems to not be set, and the contents of the `testvar` file is `\n`.

thomas11:
Resolves #48 

Address the issue that the `environment` property was seemingly ignored for remote commands.

First, don't ignore the Go error from `Setenv`.

Second, document the situation a bit better. It's complex since it depends on OS, remote shell, and SSH configuration, but hopefully we can give some additional pointers for further research.

### DIFF
--- a/provider/pkg/provider/remote/commandOutputs.go
+++ b/provider/pkg/provider/remote/commandOutputs.go
@@ -39,10 +39,13 @@ func (c *CommandOutputs) run(ctx p.Context, cmd string) error {
 	defer session.Close()
 
 	if c.Environment != nil {
-		for k, v := range *c.Environment {
+		for k, v := range c.Environment {
 			err := session.Setenv(k, v)
 			if err != nil {
-				return fmt.Errorf("could not set environment variable '%s': %w. Please see the `environment` property's documentation for guidance", k, err)
+				ctx.Logf(diag.Error, `Unable to set '%s'. This only works if your SSH server is configured to accept
+ these variables via AcceptEnv. Alternatively, if a Bash-like shell runs the command on the remote host, you could
+ prefix the command itself with the variables in the form 'VAR=value command'`, k)
+				return fmt.Errorf("could not set environment variable '%s': %w", k, err)
 			}
 		}
 	}
--- a/provider/pkg/provider/remote/commandOutputs.go
+++ b/provider/pkg/provider/remote/commandOutputs.go
@@ -40,7 +40,10 @@ func (c *CommandOutputs) run(ctx p.Context, cmd string) error {
 
 	if c.Environment != nil {
 		for k, v := range *c.Environment {
-			session.Setenv(k, v)
+			err := session.Setenv(k, v)
+			if err != nil {
+				return fmt.Errorf("could not set environment variable '%s': %w. Please see the `environment` property's documentation for guidance", k, err)
+			}
 		}
 	}
 	if c.Stdout != "" {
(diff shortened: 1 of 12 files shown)


provider/pkg/provider/remote/commandOutputs.go
```diff
@@ -40,7 +40,10 @@ func (c *CommandOutputs) run(ctx p.Context, cmd string) error {
 
 	if c.Environment != nil {
 		for k, v := range *c.Environment {
-			session.Setenv(k, v)
+			err := session.Setenv(k, v)
+			if err != nil {
+				return fmt.Errorf("could not set environment variable '%s': %w. Please see the `environment` property's documentation for guidance", k, err)
```
iwahbe, line 45: From a UX perspective, I would rather we split this up into two messages: a log message and an error:
```suggestion
				ctx.Logf(diag.Error, `Unable to set '%s'. This only works if your SSH server is configured to accept these variables via AcceptEnv. ternatively, if a Bash-like shell runs the command on the remote host, you could prefix the command itself with the variables in the form 'VAR=value command'`, k)
				return fmt.Errorf("could not set environment variable '%s': %w", k, err)
```

If you disagree, please feel free ignore this suggestion.

provider/pkg/provider/remote/commandOutputs.go
```diff
@@ -40,7 +40,10 @@ func (c *CommandOutputs) run(ctx p.Context, cmd string) error {
 
 	if c.Environment != nil {
 		for k, v := range *c.Environment {
```
iwahbe, line 42: The `CommandInputs.Environments` field shouldn't need to be a `*map[string]string`, since `map[string]string` is nillable. The change is out of scope, but I would recommend changing it to improve the authoring experience.

```suggestion
		for k, v := range c.Environment {
```

iwahbe (commented):
The change looks good. We don't *need* a test here (though it would be nice), but let's make sure we have either a test or a lint (golangci's `errcheck` would have caught this) to avoid repeating this error.

iwahbe:
You will need to regenerate SDKs.

provider/pkg/provider/remote/commandOutputs.go
thomas11: fixed manually

thomas11:
Aha. Two tests actually relied on this non-working behavior :)

```
command:remote:Command (hostname):
      error: Unable to set 'secret-key'.
```

iwahbe:
I turned on linting in ci-mgmt on this repo (https://github.com/pulumi/pulumi-command/pull/396). Looks like this would have caught the bug. ☹️

iwahbe (approved):

PR #404, opened 2024-04-16
Add flags to disable logging stdout and stderr

### LINKED ISSUE #401: stdout prints data to diff tab on pulumi up, even when an input is a secret, or stdout is marked as secret
### What happened?

I'm using `Command` to create/update/destroy some elastic policies that aren't currently supported through the official `ec` provider. 

I noticed that even when a secret is passed in as an input to the `Command` via `interpolate`, and even when `additionalSecretOutputs` is specified as `['stdout', 'stderr' ]`, the stdout of the commands are printed to the diff tab of pulumi up. 

### Expected Behavior

When I marked stdout as secret, and when I passed in a secret (I think either should do this on their own) I expected stdout to be censored, much like the commands themselves are printed as `[secret] -> [secret]`.

### Steps to reproduce

Create a trivial `Command`, such as `echo $1`, and link up a config value as input or similar. 
Such as:
```ts
const value = secret(TestCommandConfig.testValue);
new Command('test command behaviour', {
  create: interpolate`echo "${value}"`,
  update: interpolate`echo "${value}"`,
  delete: interpolate`echo "${value}"`,
}, {
    additionalSecretOutputs: ['stdout', 'stderr', 'create', 'update', 'delete']
  }
)
```
Either 1. mark the config value as a secret inline or 2. use `additionalSecretOutputs` to mark `stdout` as a secret, and then run pulumi up. 

Observe that in the first tab, the command correctly is shown as `[secret] -> [secret]` - the secret status of the input correctly propagates to the output. 
Navigate to the diff tab - observe that whatever your config value is, has been printed to the diff tab output in plain text, no censoring.

### Output of `pulumi about`

```
CLI          
Version      3.60.1
Go Version   go1.20.7
Go Compiler  gc

Host     
OS       darwin
Version  13.3.1
Arch     arm64

Backend        
Name           pulumi.com
URL            https://app.pulumi.com/instantepiphany
User           instantepiphany
Organizations  instantepiphany, displayr

Pulumi locates its logs in /var/folders/5p/ggpttgl14xxf5k01pl3rd6x40000gn/T/nix-shell.aVRQAW by default
warning: Failed to read project: no Pulumi.yaml project file found (searching upwards from /Users/jordan/Source/displayr/dip). If you have not created a project yet, us
e `pulumi new` to do so: no project file found
warning: Failed to get information about the current stack: no Pulumi.yaml project file found (searching upwards from /Users/jordan/Source/displayr/dip). If you have no
t created a project yet, use `pulumi new` to do so: no project file found
warning: A new version of Pulumi is available. To upgrade from version '3.60.1' to '3.80.0', visit https://pulumi.com/docs/reference/install/ for manual instructions an
d release notes.
jordan@Jordans-MBP dip % 
```

### Additional context

That the secret exposure is only visible on the diff tab, and not on the main tab (or on the local terminal if pulumi up was run locally) makes this worse - I was not aware of this exposure until a colleague found it, and I then found other instances of this after reviewing usage of `Command`.

### Contributing

thomas11:
Resolves #401 

This PR adds a flag to both the local and the remote Command resource, allowing the user to opt out of logging the command's output. This is useful when the command may contain secrets, or when it's simply too verbose.

Implementation notes:
- The test for the remote Command uses `github.com/gliderlabs/ssh` to spin up an in-process SSH server to be able to unit test command execution.
- I extracted some of the repeated input property definitions into `common/inputs.go`.


provider/cmd/pulumi-resource-command/schema.json
```diff
@@ -217,6 +217,11 @@
           },
           "description": "The program and arguments to run the command.\nOn Linux and macOS, defaults to: `[\"/bin/sh\", \"-c\"]`. On Windows, defaults to: `[\"cmd\", \"/C\"]`"
         },
+        "logOutput": {
+          "type": "boolean",
+          "description": "If the command's stdout and stderr should be logged.",
```
flostadler, line 222: I wonder if users might wanna select what streams end up being logged, e.g. stderr only.

By using a boolean we kind of limit the API for the future

provider/pkg/provider/local/command.go
```diff
@@ -42,29 +43,15 @@ func (c *Command) Annotate(a infer.Annotator) {
...
 // These are the inputs (or arguments) to a Command resource.
 type CommandInputs struct {
-	// The field tags are used to provide metadata on the schema representation.
-	// pulumi:"optional" specifies that a field is optional. This must be a pointer.
-	// provider:"replaceOnChanges" specifies that the resource will be replaced if the field changes.
-	Triggers *[]any  `pulumi:"triggers,optional" provider:"replaceOnChanges"`
-	Create   *string `pulumi:"create,optional"`
-	Update   *string `pulumi:"update,optional"`
-	Delete   *string `pulumi:"delete,optional"`
+	common.ResourceInputs
```
iwahbe, line 46: That this "just works" makes me unreasonably happy.

provider/pkg/provider/util/util.go
```diff
@@ -26,14 +28,20 @@ import (
...
+func LogOutput(ctx p.Context, r io.Reader, doneCh chan<- struct{}, severity diag.Severity) {
 	defer close(doneCh)
 	scanner := bufio.NewScanner(r)
 	for scanner.Scan() {
 		ctx.LogStatus(severity, scanner.Text())
 	}
 }
 
+// NoopLogger is for testing. It reads from the provided reader until EOF, discarding the output, then closes the channel.
+func NoopLogger(r io.Reader, done chan struct{}) {
```
iwahbe, line 40: It seems like we use this outside of testing.

provider/pkg/provider/util/util.go
```diff
@@ -44,3 +52,19 @@ func (w *ConcurrentWriter) Write(bs []byte) (int, error) {
...
+
+func (c *TestContext) log(msg string) {
+	c.Output.WriteString(msg)
+}
+
+func (c *TestContext) Log(_ diag.Severity, msg string)                  { c.log(msg) }
+func (c *TestContext) Logf(_ diag.Severity, msg string, _ ...any)       { c.log(msg) }
+func (c *TestContext) LogStatus(_ diag.Severity, msg string)            { c.log(msg) }
+func (c *TestContext) LogStatusf(_ diag.Severity, msg string, _ ...any) { c.log(msg) }
+func (c *TestContext) RuntimeInformation() p.RunInfo                    { return p.RunInfo{} }
```
iwahbe, line 70: This shouldn't be accessible as part of the non-test code. Can we move it to `util/util_test.go` so it is only usable from tests?

provider/pkg/provider/remote/command.go
```diff
@@ -29,38 +30,28 @@ The connection is established via ssh.`)
...
-	Delete      *string           `pulumi:"delete,optional"`
-	Update      *string           `pulumi:"update,optional"`
-	Stdin       *string           `pulumi:"stdin,optional"`
 }
 
 // Implementing Annotate lets you provide descriptions and default values for arguments and they will
 // be visible in the provider's schema and the generated SDKs.
 func (c *CommandInputs) Annotate(a infer.Annotator) {
+	c.ResourceInputs.Annotate(a)
+	c.CommonInputs.Annotate(a)
```
iwahbe, line 48: Is this necessary? It shouldn't be.

provider/pkg/provider/remote/command.go
```diff
@@ -15,6 +15,7 @@
 package remote
 
 import (
+	"github.com/pulumi/pulumi-command/provider/pkg/provider/common"
 	"github.com/pulumi/pulumi-go-provider/infer"
```
iwahbe, line 19:
```suggestion
	"github.com/pulumi/pulumi-go-provider/infer"
	
	"github.com/pulumi/pulumi-command/provider/pkg/provider/common"
```

iwahbe (commented):
I agree with @flostadler on the API design. Maybe we should make it an enum so we can add "stdout-only" and "stderr-only" variants at need.

thomas11:
> I agree with @flostadler on the API design. Maybe we should make it an enum so we can add "stdout-only" and "stderr-only" variants at need.

Smells like YAGNI to me, but I'll give it a shot.

provider/pkg/provider/util/util.go
thomas11: True, thanks

Issue #196, opened 2023-05-02
Add examples in all languages


lukehoban:
Two pieces of work:
- [x] In the README, include copies of all examples in all languages (currently several are TS-only). https://www.pulumi.com/registry/packages/command/
- [x] Add per-resource examples in the Schema in all languages to get projected into per-resource pages like https://www.pulumi.com/registry/packages/command/api-docs/local/command/.

PR #423, opened 2024-04-30
CopyFile: support assets, archives, and recursive copying

### LINKED ISSUE #23: Expand the scope of remote:CopyFile to support copying directories

### LINKED ISSUE #33: Consider contents of file for changes in CopyFile

### LINKED ISSUE #42: Add support for Asset as argument to CopyFile

thomas11:
This PR allows the [CopyFile](https://www.pulumi.com/registry/packages/command/api-docs/remote/copyfile/) resource to recursively copy directories as well, similar to `scp -r`.

This PR enhances the CopyFile resource in a few ways.
- It can now copy directories as well, recursively with their contents.
- It now takes the standard Pulumi [asserts and archives](https://www.pulumi.com/docs/concepts/assets-archives/) as input, allowing for seamless interop with other resources.
- It now checks whether the specified file or directory have changed (in content, not timestamp) and copies only if they did.

In light of these changes, I've renamed `CopyFile` to just `Copy` which is a **breaking change**.

Resolves #23
Resolves #33
Resolves #42

## TODO

- [x] ~~The current implementation fails as soon as a file or directory already exists on the remote, like the previous `CopyFile`.~~
- [x] Before GA, should we rename `CopyFile` to `Copy` or `RemoteCopy` or _insert your proposal here_?
- [ ] The integration test copies a bunch of EC2 setup code with `TestEc2RemoteTs`, which we should share instead. _What would be a good way to do that, given the code is in the TS tests, not in the Go driver?_

## Design considerations

The behavior of the copy operation was modeled after `cp` and `scp`.

```
source | dest - exists as dir | dest - does not exist | dest - exists as file
-------|----------------------|-----------------------|-----------------------
dir    | dest/dir             | dest/dir              | error
dir/   | dest/x for x in dir  | dest/dir              | error
file   | dest/file            | dest                  | dest (overwritten)
```

Specifically:
- When copying a directory, we overwrite existing files. (The current CopyFile resource for single files does that, too.)
- When copying a directory, we _don't_ clear the remote directory first so that no left-over files from previous copies will be around. We can always add a flag for that if needed.

## Implementation notes

- I've looked around for open source Go packages implementing scp of folders, but to my surprise, I couldn't find one with an acceptable license that 1) wasn't ancient and 2) was cross-platform.
- The current implementation copies files sequentially and is therefore probably slow for large trees. If we replaced the implementation, I don't think the shape of the resource would change, so this shouldn't be a GA blocker.
- The `size.ts` file in both steps of the integration test is useless because it always has the same value. However, I found that without a TS file present in the additional step, it would not be run. Haven't debugged further.


provider/pkg/provider/remote/copy.go
```diff
@@ -0,0 +1,82 @@
...
+	}
+	if c.LocalAsset == nil && c.LocalArchive == nil {
+		return errors.New("either LocalAsset or LocalArchive must be set")
+	}
+	if c.LocalAsset != nil && !c.LocalAsset.IsPath() {
+		return errors.New("LocalAsset must be a file asset")
+	}
+	if c.LocalArchive != nil && !c.LocalArchive.IsPath() {
+		return errors.New("LocalArchive must be a path to a file or directory")
+	}
```
iwahbe, line 62: Do we need this restriction? Can we support an arbitrary asset or archive? 

We can relax this restriction after GA if we can't get it in time.
If we can only support file-path based assets and archives we need to document that in the field descriptions explicitly:

"An asset to upload. The asset must be a path based asset."

provider/pkg/provider/remote/copy.go
```diff
@@ -0,0 +1,82 @@
...
+
+func (c *CopyInputs) Annotate(a infer.Annotator) {
+	a.Describe(&c.Connection, "The parameters with which to connect to the remote host.")
+	a.Describe(&c.Triggers, "Trigger replacements on changes to this input.")
+	a.Describe(&c.LocalAsset, "The path of the file to be copied. Only one of LocalAsset or LocalArchive can be set.")
+	a.Describe(&c.LocalArchive, "The path of the folder or archive to be copied. Only one of LocalAsset or LocalArchive can be set.")
+	a.Describe(&c.RemotePath, "The destination path in the remote host.")
+}
+
+func (c *CopyInputs) validate() error {
```
iwahbe, line 50: This has a couple problems as implemented:

1. It doesn't run on new resources, only on updates.
2. I'm not confident it works correctly for unknown assets and archives during preview.

We can fix both of these problems by implementing `Check`, instead of dancing around it. Lets actually implement `Check`, do the "ensure exclusivity of fields" on the raw property map (so unknowns are handled correctly), call `infer.DefaultCheck` to serialize to `CopyInputs`, then do "IsPath()" check.

provider/pkg/provider/remote/copyController.go
```diff
@@ -0,0 +1,153 @@
...
+		return olds, err
+	}
+
+	needCopy := true
+	if news.hash() == olds.hash() && news.RemotePath == olds.RemotePath {
+		needCopy = false
+	}
+
+	if preview {
+		// TODO,tkappler how to show the user whether we need to copy or not?
```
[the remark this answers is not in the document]
thomas11: I was thinking of preview, since we're inside the `if preview` block. Shouldn't the preview show the user what will be done? I.e., will there be a fresh copy?

provider/pkg/provider/remote/copy.go
```diff
@@ -0,0 +1,82 @@
...
+// Copy implements Annotate which allows you to attach descriptions to the Copy resource.
+func (c *Copy) Annotate(a infer.Annotator) {
+	a.Describe(&c, "Copy a local file to a remote host.")
+}
+
+type CopyInputs struct {
+	Connection   *Connection      `pulumi:"connection" provider:"secret"`
+	Triggers     *[]interface{}   `pulumi:"triggers,optional" providers:"replaceOnDelete"`
+	LocalAsset   *asset.Asset     `pulumi:"localAsset,optional"`
+	LocalArchive *archive.Archive `pulumi:"localArchive,optional"`
```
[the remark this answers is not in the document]
thomas11: Any particular reason? I prefer the added clarity at the cost of verbosity.

provider/pkg/provider/remote/copy.go
thomas11: Great point. Even though I theoretically knew about `Check`, it wasn't on my radar because I'd never worked with it.

provider/pkg/provider/remote/copy.go
iwahbe: > I haven't fully thought through what support for non-path-based assets/archives means and was playing it safe to get started.

That makes sense. I'm OK with delaying this until after GA.

> What should the resource do with a

> * StringAsset: write the string into the remote file?

Yes

> * RemoteAsset: download first, then handle like a path asset?

That's what I would do. If we wanted to be fancier (and more efficient), we could SSH into the target system and initiate the copy from their (avoiding the relay), but that could introduce more failure types. I would stick to download then copy.

> * RemoteArchive: same

Same

> * AssetArchive: recurse through and do one of the above for each one?

Yep

provider/pkg/provider/remote/copy.go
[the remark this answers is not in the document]
iwahbe: It's related to https://github.com/pulumi/pulumi-command/pull/423#discussion_r1597285198. Changing these names is breaking, and I don't think we want to always require a local asset to be used.

When users use a remote asset (say in typescript), then they would need to pass 

```ts
  localAsset: new pulumi.asset.RemoteAsset(remoteURL),
```

Since we will want to expand the `LocalAsset` to accept all kinds of Pulumi assets before v2, I would rather users write:

```ts
  asset: new pulumi.asset.RemoteAsset(remoteURL),
```

provider/pkg/provider/remote/copyController.go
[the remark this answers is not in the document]
iwahbe: I'm concerned it will be noisy, but I might be wrong. As I understand it, it's just adding or removing a couple of lines, so I would implement the logging and see if its too noisy. If you like the more verbose messaging better, then leave it in.

We can futz with the logging after GA, so I'll be happy with whatever you think best.

provider/pkg/provider/remote/copy.go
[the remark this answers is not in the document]
thomas11: Makes sense. I'm just a little concerned about clarity regarding the "from" and "to" of the copy. The "to" is always remote, but the "from" might also be when we open up to all kinds of assets. But `FromAsset` is not a great name either.

provider/pkg/provider/remote/copy.go
thomas11: Tracked in https://github.com/pulumi/pulumi-command/issues/431

provider/pkg/provider/remote/copy.go
```diff
@@ -30,33 +30,33 @@ func (c *Copy) Annotate(a infer.Annotator) {
...
+	RemotePath string           `pulumi:"remotePath"`
 }
 
 func (c *CopyInputs) Annotate(a infer.Annotator) {
 	a.Describe(&c.Connection, "The parameters with which to connect to the remote host.")
 	a.Describe(&c.Triggers, "Trigger replacements on changes to this input.")
-	a.Describe(&c.LocalAsset, "An asset to upload. It must be a path based asset. Only one of LocalAsset or LocalArchive can be set.")
-	a.Describe(&c.LocalArchive, "An archive to upload. It must be a path based archive. Only one of LocalAsset or LocalArchive can be set.")
+	a.Describe(&c.Asset, "An asset to upload as the source of the copy. It must be a path based asset. Only one of LocalAsset or LocalArchive can be set.")
+	a.Describe(&c.Archive, "An archive to upload as the source of the copy. It must be a path based archive. Only one of LocalAsset or LocalArchive can be set.")
```
iwahbe, line 44: It's not LocalAsset and LocalArchive anymore.

provider/pkg/provider/remote/copyController.go
```diff
@@ -0,0 +1,153 @@
...
+		return err
+	}
+	defer remote.Close()
+
+	_, err = remote.ReadFrom(local)
+	return err
+}
+
+// copyDir copies a directory recursively from the local file system to a remote host.
+// Note that the current is naive and sequential and therefore can be slow.
```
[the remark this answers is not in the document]
flostadler: We should also think about turning on concurrent writes/reads if we're having performance issues: https://github.com/pkg/sftp/blob/v1.13.6/client.go#L109

That's what the sftp linux tool is doing as well. I saw some rather big performance improvements when using that with the go library (more than 2x), but it makes error handling more complex

provider/pkg/provider/remote/copyController.go
```diff
@@ -0,0 +1,195 @@
...
+		dirInfo, err := sftp.Stat(remotePath)
+		// sftp normalizes the error to os.ErrNotExist, see client.go: normaliseError.
+		if err != nil && !errors.Is(err, os.ErrNotExist) {
+			return err
+		}
+
+		if dirInfo == nil {
+			return sftp.Mkdir(remotePath)
+		} else if !dirInfo.IsDir() {
+			err = sftp.Remove(remotePath)
```
flostadler, line 188: Don't we still need to create the remote directory after deleting the file with the same name?

provider/pkg/provider/remote/copyController.go
```diff
@@ -0,0 +1,195 @@
...
+		}
+
+		dirInfo, err := sftp.Stat(remotePath)
+		// sftp normalizes the error to os.ErrNotExist, see client.go: normaliseError.
+		if err != nil && !errors.Is(err, os.ErrNotExist) {
+			return err
+		}
+
+		if dirInfo == nil {
+			return sftp.Mkdir(remotePath)
```
flostadler, line 186: What if the parent of `dst` doesn't exist? This would end up failing, right? Should we maybe use `MkdirAll` here to be safe? Or maybe better do that in the `doCopy` function because `copyFile` needs it to, right?

provider/pkg/provider/remote/copyController.go
```diff
@@ -0,0 +1,195 @@
...
+}
+
+func copyFile(sftp *sftp.Client, src, dst string) error {
+	local, err := os.Open(src)
+	if err != nil {
+		return err
+	}
+	defer local.Close()
+
+	remote, err := sftp.Create(dst)
```
flostadler, line 154: What if the parents of dst do not exist?

examples/ec2_dir_copy/index.ts
```diff
@@ -0,0 +1,93 @@
...
+            { protocol: "tcp", fromPort: 80, toPort: 80, cidrBlocks: ["0.0.0.0/0"] },
+        ],
+    });
+
+    const ami = aws.ec2.getAmiOutput({
+        owners: ["amazon"],
+        mostRecent: true,
+        filters: [{
+            name: "name",
+            values: ["amzn2-ami-hvm-*-x86_64-ebs"],
```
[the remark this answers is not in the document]
thomas11: Thanks for the hint, updated

provider/pkg/provider/remote/copyController.go
thomas11: When `copyFile` is called from our `copyDir`, that shouldn't be possible.

When it's called directly because the user wants to copy a file, one could argue that it should fail because the user asked to copy to a non-existant location. It's one of the questions in my (internal) ideation doc.

provider/pkg/provider/remote/copyController.go
thomas11: In a recursive call, that shouldn't be possible.

In the first call - creating the user's top-level directory - one could argue that it should fail because the user asked to copy to a non-existent location. It's one of the questions in my (internal) ideation doc.

provider/pkg/provider/remote/copyController.go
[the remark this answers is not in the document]
thomas11: Turned on concurrent writes in the latest commit. Concurrent reads are already on by default.

provider/pkg/provider/remote/copyController.go
```diff
@@ -0,0 +1,246 @@
...
+			Property: "asset",
+			Reason:   "either asset or archive must be set",
+		})
+	}
+
+	inputs, newFailures, err := infer.DefaultCheck[CopyInputs](newInputs)
+	if err != nil {
+		return inputs, failures, err
+	}
+	failures = append(failures, newFailures...)
```
iwahbe, line 62: Regardless of if there is an error, we should merge `failures` and `newFailures`.

```suggestion
	inputs, newFailures, err := infer.DefaultCheck[CopyInputs](newInputs)
	failures = append(failures, newFailures...)
	if err != nil {
		return inputs, failures, err
	}
```

iwahbe (approved):
This looks pretty good to me.

provider/pkg/provider/remote/copy.go
```diff
@@ -0,0 +1,64 @@
...
+// Copy implements Annotate which allows you to attach descriptions to the Copy resource.
+func (c *Copy) Annotate(a infer.Annotator) {
+	a.Describe(&c, "Copy an Asset or Archive to a remote host.")
+}
+
+type CopyInputs struct {
+	Connection *Connection      `pulumi:"connection" provider:"secret"`
+	Triggers   *[]interface{}   `pulumi:"triggers,optional" providers:"replaceOnDelete"`
+	Asset      *asset.Asset     `pulumi:"asset,optional"`
+	Archive    *archive.Archive `pulumi:"archive,optional"`
```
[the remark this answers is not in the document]
thomas11: Oooh interesting! I just looked at the definitions of `Asset` and `Archive` and saw that they're separate.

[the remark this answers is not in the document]
danielrbradley: Yeah, the design of Assets overall is pretty confusing on both the provider and consumer sides 🤷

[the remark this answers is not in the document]
danielrbradley: Here's an example from azure-native:

https://github.com/pulumi/pulumi-azure-native/blob/f607b7a1e27913aef9672b75913bdf341c46c118/provider/pkg/resources/customresources/custom_storage.go#L659-L678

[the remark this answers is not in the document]
danielrbradley: This works a bit differently because it's using raw property values rather than the new framework. Perhaps this is a missing feature in the go-provider to detect which kind of asset we've got at runtime?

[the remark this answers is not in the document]
thomas11: This discussion led to filing https://github.com/pulumi/pulumi-go-provider/issues/237

provider/cmd/pulumi-resource-command/schema.json
```diff
@@ -469,22 +469,22 @@
         "connection"
       ]
     },
-    "command:remote:CopyFile": {
```
danielrbradley, line 472: Should we include an alias for this rename? Probably not too significant because this resource doesn't really use state in any meaningful way.

provider/pkg/provider/remote/copy.go
```diff
@@ -0,0 +1,61 @@
...
+	"github.com/pulumi/pulumi-go-provider/infer/types"
+)
+
+type Copy struct{}
+
+var _ = (infer.Annotated)((*Copy)(nil))
+
+// Copy implements Annotate which allows you to attach descriptions to the Copy resource.
+func (c *Copy) Annotate(a infer.Annotator) {
+	a.Describe(&c, "Copy an Asset or Archive to a remote host.")
```
danielrbradley, line 28: Would be nice to really bulk out this documentation with examples etc.

Could be a follow-up change though.

provider/pkg/provider/remote/copy.go
```diff
@@ -0,0 +1,61 @@
...
+	Connection *Connection          `pulumi:"connection" provider:"secret"`
+	Triggers   *[]interface{}       `pulumi:"triggers,optional" provider:"replaceOnChanges"`
+	Source     types.AssetOrArchive `pulumi:"source"`
+	RemotePath string               `pulumi:"remotePath"`
+}
+
+func (c *CopyInputs) Annotate(a infer.Annotator) {
+	a.Describe(&c.Connection, "The parameters with which to connect to the remote host.")
+	a.Describe(&c.Triggers, "Trigger replacements on changes to this input.")
+	a.Describe(&c.Source, "An asset or an archive to upload as the source of the copy. It must be path based.")
```
danielrbradley, line 41: What does "it must be path based" mean here? Does this fail if someone passes in a StringAsset and sets `RemotePath` to "~/test.txt"?

danielrbradley: Might also be nice to include specific examples of specific asset and archive types and how they'll be transferred.
- Does an AssetArchive get unpacked at the other end or does it depend on the remotePath? If it's remote path is `~/dest.tar.gz` does it keep it packed?
- Does an Asset assume the remote path is a filename and an archive assume it's a directory?

provider/cmd/pulumi-resource-command/schema.json
thomas11: I need to do some testing around this. It's not only the name of the resource that changes, the main input property for the source of the copy changes as well. I think a replace might be unavoidable.

provider/pkg/provider/remote/copy.go
thomas11: Yep, this is tracked by #196 and will be a separate PR.

provider/cmd/pulumi-resource-command/schema.json
danielrbradley: I think you'll be okay as that property doesn't cause replacements.

Issue #445, opened 2024-05-17
pulumi_command.common has no attribute Logging


bonovoxly:
### What happened?

Using the latest pulumi command returns this error (just started happening this morning):

```
 AttributeError: module 'pulumi_command.common' has no attribute 'Logging'
```

### Example

some snippets:

```
from pulumi_command import local
```

```
        self.local_command = local.Command(f"{self.lambda_name}_zip", 
                                           create=command_zip,
                                           update=command_zip,
                                           triggers=[{"contentHash": content_hash}],
                                           asset_paths=[f"{self.src_path}/{self.lambda_name}.zip"]
                                           )
```

### Output of `pulumi about`

```
pulumi about  
CLI          
Version      3.112.0
Go Version   go1.22.1
Go Compiler  gc

Plugins
NAME        VERSION
aws         5.43.0
aws-iam     0.2.0
command     0.11.0
kubernetes  4.11.0
python      unknown

Host     
OS       darwin
Version  14.5
Arch     arm64
```

### Additional context

verified no error in `0.10.0`.

### Contributing

Vote on this issue by adding a 👍 reaction. 
To contribute a fix for this issue, leave a comment (and link to your pull request, if you've opened one already).

thomas11:
Hi @bonovoxly, thank you for reporting this. I can see a weird change in the Python SDK in v0.11.0 and will investigate. Are you able to stay on v0.10.0 in the meantime?

bonovoxly:
@thomas11  yeah we can. just wanted to make sure that IF it was a bug, someone was notified. 

(I was secretly hoping it was `user error` ;) )

iwahbe:
This is due to https://github.com/pulumi/pulumi/issues/16221.

thomas11:
Fixed in v0.11.1.

PR #448, opened 2024-05-17
Move the Logging type to work around Python codegen issue 16221

thomas11:
The `Logging` enum defined in the `common` package was not correctly exposed in the Python SDK due to pulumi/pulumi#16221, causing the error 
```
AttributeError: module 'pulumi_command.common' has no attribute 'Logging'
```

As a workaround, this PR moves the Logging enum into both the `local` and the `remote` package. This duplicates some code but keeps the packages separated. As the provider is not GA yet, we could reverse this once the upstream is fixed, at the cost of a breaking change.

Tested a local `pulumi up`.

Fixes https://github.com/pulumi/pulumi-command/issues/445


iwahbe (approved):

lunaris:
As discussed in Slack my recommendation would be to duplicate the `Logging` type into `local` and `remote`, but happy for us to go another way!

PR #355, opened 2024-03-08
Allow opt-out of adding stdout/stderr env variables

julsemaan:
fixes #285

PR #451, opened 2024-05-24
Allow opt-out of adding stdout/stderr env variables for remote command

### LINKED ISSUE #285: error: fork/exec /bin/sh: argument list too long
### What happened?

Running `pulumi up` on an existing stack on Linux fails.

Poking at the code, I see this is because of PULUMI_COMMAND_STDOUT/PULUMI_COMMAND_STDERR and the previous run containing rather long output.

https://github.com/pulumi/pulumi-command/blob/7c5c4bd68e66950c042c5ce1ad3eb6f9bfda835c/provider/pkg/provider/local/commandOutputs.go#L73-L78

Since I don't use this feature and I don't always have the option of adjusting limits, I'd like an option to turn it off. Furthermore, I'd prefer if the option could be made off by default as well, as I don't generally need this information in state files.

### Example

strace
```
sudo -E strace -E PATH=$PATH -u $USER -f -e trace=execve -s 1000 -v --signals=SIGPWR pulumi up 2> log
```

From log above, we can tell our command is not particularly long, the environment is massive.
```
execve("/bin/sh", ["/bin/sh", "-c", "ansible REDACTED"], ["SHELL=/bin/bash", ...REDACTED..., "PULUMI_COMMAND_STDOUT=ansible-playbook [core 2.14.6]\n  config file REDACTED"...]) = -1 E2BIG (Argument list too long)
```

### Output of `pulumi about`

```
CLI
Version      3.75.0
Go Version   go1.20.5
Go Compiler  gc

Host
OS       ubuntu
Version  20.04
Arch     x86_64

Backend
Name           REDACTED
URL            file://~
User           REDACTED
Organizations
```

### Additional context

_No response_

### Contributing

julsemaan:
# Context

At the moment, stdout and stderr of previous runs are passed as environment variables on future invocations. When stdout or stderr is too large, this cause an error. 

# Proposed change

#355 fixed it for local commands, this fixed it for remote commands. See #285 for full details on the issue

See this comment for details around local vs remote functionality : https://github.com/pulumi/pulumi-command/issues/285#issuecomment-2122415592


julsemaan:
I wasn't able to successfully write a test for this one given I couldn't use the preview mode and it was trying to actually SSH into hosts. I'm lacking understanding of how to properly mock this within this plugin. Not sure if the PR can still be merged without the test.

thomas11:
Thanks for the PR @julsemaan!

The change is simple enough that we could get away without a test. But if you wanted to try your hand at it, I've had success with spinning up a local, in-process SSH server and testing that way. The PR isn't merged yet but you can see it [here](https://github.com/pulumi/pulumi-command/pull/423/files#diff-f3e686af89272926b32943bc7accc9653687755256bfdb4247761615bace2f90).

julsemaan:
Hey! If I can get away without a test, I'll take it 😄 

I could always implement a test in a follow-up (pinky promess?). I'll be able to reuse some of your code vs doing some copy/pasta

thomas11:
Sure, I think we can get away with that :)

However, I just noticed that your change is missing adding the new property to `Annotate` for the description and setting the default to true.

provider/pkg/provider/remote/command.go
```diff
@@ -55,6 +56,10 @@ outputs as secret via 'additionalSecretOutputs'. Defaults to logging both stdout
 Note that this only works if the SSH server is configured to accept these variables via AcceptEnv.
 Alternatively, if a Bash-like shell runs the command on the remote host, you could prefix the command itself
 with the variables in the form 'VAR=value command'.`)
+	a.Describe(&c.AddPreviousOutputInEnv,
+		`If the previous command's stdout and stderr (as generated by the prior create/update) is
+injected into the environment of the next run as PULUMI_COMMAND_STDOUT and PULUMI_COMMAND_STDERR.
+Defaults to true.`)
 }
```
thomas11, line 63:
```suggestion
	a.SetDefault(&c.AddPreviousOutputInEnv, true)
}
```

thomas11 (changes requested):

mikhailshilkov:
@julsemaan For any current reviewers and future readers, could you please update the description of your PR to include a detailed explanation of your change and motivation for doing it?

Here are some example of good PR descriptions: https://github.com/pulumi/pulumi-kubernetes-operator/pull/575, https://github.com/pulumi/pulumi-github/pull/280, https://github.com/pulumi/pulumi/pull/13066.

Thank you for your contribution!

julsemaan:
> @julsemaan For any current reviewers and future readers, could you please update the description of your PR to include a detailed explanation of your change and motivation for doing it?
> 
> Here are some example of good PR descriptions: [pulumi/pulumi-kubernetes-operator#575](https://github.com/pulumi/pulumi-kubernetes-operator/pull/575), [pulumi/pulumi-github#280](https://github.com/pulumi/pulumi-github/pull/280), [pulumi/pulumi#13066](https://github.com/pulumi/pulumi/pull/13066).
> 
> Thank you for your contribution!

Hey! That's done, I could suggest to have a PR template so that contributors know more precisely what is expected to provide when opening a PR. Github makes this quite easy, see https://docs.github.com/en/communities/using-templates-to-encourage-useful-issues-and-pull-requests/creating-a-pull-request-template-for-your-repository

thomas11:
@julsemaan, on second thought, I believe we should include test coverage here. If you just copy roughly lines 23-53 from [here](https://github.com/pulumi/pulumi-command/pull/423/files#diff-f3e686af89272926b32943bc7accc9653687755256bfdb4247761615bace2f90R23), that should give you the setup to use SSH in your test.

theneva:
Hi, @julsemaan! Thank you so much for implementing the fix :smile:

Given that this whole thing stems from my feature request, and you never asked to dive into the remote command resource, I went ahead and wrote a test that I think does the job. I sent a pull request to your fork here: https://github.com/julsemaan/pulumi-command/pull/1

@thomas11 Do you think the test I wrote looks okay? (If yes, I believe you are also able to pull my commit into this PR, assuming Julien turned maintainer edits on.)

julsemaan:
@theneva , thanks for doing the test :) I just pulled it into this branch

provider/tests/provider_test.go
```diff
@@ -223,6 +226,123 @@ func TestRemoteCommand(t *testing.T) {
...
+				"connection":             connection,
+				"create":                 resource.NewStringProperty(createCommand),
+				"addPreviousOutputInEnv": resource.NewBoolProperty(addPreviousOutputInEnv),
+			},
+		})
+		require.NoError(t, err)
+		return resp.Properties
+	}
+
+	t.Run("create-actual", func(t *testing.T) {
```
thomas11, line 318: `create()` and `createdState` are mixed in with the other bits, but they are more like test setup, right? The three sub-tests need to run in order for this to work?

Maybe it would be nicer and less error-prone (when someone turns on t.Parallel() to extract all setup into a method and call it from both update sub-tests?

provider/tests/provider_test.go
```diff
@@ -223,6 +226,123 @@ func TestRemoteCommand(t *testing.T) {
...
+			"addPreviousOutputInEnv": resource.PropertyValue{V: true},
+		}, update(true))
+	})
+
+	t.Run("update-actual-without-std", func(t *testing.T) {
+		assert.Equal(t, resource.PropertyMap{
+			"connection": connection,
+			"create":     resource.PropertyValue{V: createCommand},
+			"stderr":     resource.PropertyValue{V: ""},
+			// Running with addPreviousOutputInEnv does not set the environment variable.
```
thomas11, line 338: "Running without"

theneva:
@thomas11 Thanks for the review! I tried following your feedback here: https://github.com/julsemaan/pulumi-command/pull/2/files.

Since I don't have push access to the main repo _or_ Julien's fork, this workflow kind of sucks for everyone involved… I think it's easiest if you just look at the PR I sent to Julien's branch in case you have more feedback, so that they don't have to jump for every feedback loop.

(Sorry for all the notifications, @julsemaan 😅)

thomas11 (approved):

thomas11:
Merged! Thank you very much for all your efforts, @julsemaan and @theneva!

PR #452, opened 2024-05-24
Correct struct tag and property name of Triggers replace annotation

thomas11:
The struct tag pulumi-go-provider looks for is `provider`, not `providers`.

Also, `replaceOnDelete` is not a thing. Based on the property docs "Trigger replacements on changes to this input" I assume replace on changes is meant. Note that two integration tests already use `Triggers` and pass. I assume this is because a change to `Triggers` causes not a replacement of the resource but a fresh copy?

Plus a bonus docs fix for Connection.Port.

### DIFF
--- a/provider/pkg/provider/remote/copyfile.go
+++ b/provider/pkg/provider/remote/copyfile.go
@@ -29,7 +29,7 @@ func (c *CopyFile) Annotate(a infer.Annotator) {
 
 type CopyFileInputs struct {
 	Connection *Connection    `pulumi:"connection" provider:"secret"`
-	Triggers   *[]interface{} `pulumi:"triggers,optional" providers:"replaceOnDelete"`
+	Triggers   *[]interface{} `pulumi:"triggers,optional" provider:"replaceOnChanges"`
 	LocalPath  string         `pulumi:"localPath"`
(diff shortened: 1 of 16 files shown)


provider/pkg/provider/remote/copyfile.go
```diff
@@ -29,7 +29,7 @@ func (c *CopyFile) Annotate(a infer.Annotator) {
 
 type CopyFileInputs struct {
 	Connection *Connection    `pulumi:"connection" provider:"secret"`
-	Triggers   *[]interface{} `pulumi:"triggers,optional" providers:"replaceOnDelete"`
+	Triggers   *[]interface{} `pulumi:"triggers,optional" provider:"replaceOnChanges"`
```
iwahbe, line 32: Should we look for this and error during schema generation?

I don't know if it's idiomatic but typos are painful.

iwahbe (approved):

provider/pkg/provider/remote/copyfile.go
thomas11: Is there an existing way to warn, not error, in the code base already? I couldn't see anything.

PR #470, opened 2024-06-18
chore: make the optional logging test use the common SSH server

theneva:
@thomas11 as promised in #451, I migrated `TestOptionalLogging(…)` to use the SSH server abstraction.

### DIFF
--- a/provider/pkg/provider/remote/commandController_test.go
+++ b/provider/pkg/provider/remote/commandController_test.go
@@ -28,8 +28,15 @@ import (
 )
 
 func TestOptionalLogging(t *testing.T) {
-	for _, logMode := range Logging.Values(LogStdoutAndStderr) {
+	for _, sharedLogMode := range Logging.Values(LogStdoutAndStderr) {
+		// Get a copy of the log mode that doesn't get changed by the next iteration in the for loop,
+		// to allow running in parallel.
+		// See 'The long-standing "for" loop gotcha' here: https://go.dev/blog/go1.22
+		// This can be removed after upgrading to go 1.22.
+		logMode := sharedLogMode
 		t.Run(logMode.Name, func(t *testing.T) {
+			t.Parallel()
+
 			// This SSH server always writes "foo" to stdout and "bar" to stderr, no matter the command.
 			server := testutil.NewTestSshServer(t, func(s ssh.Session) {
 				_, err := io.WriteString(s, "foo")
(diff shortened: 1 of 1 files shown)


provider/pkg/provider/remote/commandController_test.go
```diff
@@ -29,31 +28,16 @@ import (
...
-
 	for _, logMode := range Logging.Values(LogStdoutAndStderr) {
 		t.Run(logMode.Name, func(t *testing.T) {
+			// This SSH server always writes "foo" to stdout and "bar" to stderr, no matter the command.
+			server := testutil.NewTestSshServer(t, func(s ssh.Session) {
+				_, err := io.WriteString(s, "foo")
+				require.NoError(t, err)
+				_, err = io.WriteString(s.Stderr(), "bar")
+				require.NoError(t, err)
+			})
```
theneva, line 39: I had to move the server creation inside the subtest, since it depends on `t` to fail only the appropriate test if something errors inside the handler.

I'm not worried about port conflicts, because the server requests port 0, so each instance gets a free port.

**HOWEVER.** I was certain that this isolation of each test would also let me add `t.Parallel()` to the beginning of the function. It does not.

If I add `t.Parallel()`, the tests get reported as passing even if I change the strings that are written. Do you have any idea why that's happening, @thomas11?

theneva: … can it be because the _commands_ that are created all have the same `name`, and therefore operate on the same resource?

thomas11: That's quite surprising. Even with a shared SSH server it should work in parallel since the server has no state.

Output is captured into `ctx.Output` which is also recreated for each test, right?

theneva: … oh, I see what's happening: It's the [long-standing “for” loop gotcha](https://go.dev/blog/go1.22), which they finally changed in go 1.22 – but this project is on 1.21.

Effectively, it causes every test to read the last `logMode` entry in the slice returned by `Logging.Values(…)`,  because the for loop finishes its iterations before the tests read `logMode.ShouldLogStdXXX()`.

Changing `go 1.21` to `go 1.22` in `provider/go.mod` immediately fixes the problem. Is upgrading to 1.22 a scary or difficult change?

I'll work around it in the test – one second

theneva: This commit works around the issue on go 1.21: 1a3b83dced24107a36dbb436c5ace4462109b59c

thomas11 (approved):

PR #472, opened 2024-06-20
Document Command in its own Markdown file, with `triggers` examples

### LINKED ISSUE #40: Document functionality of triggers property on Command resource
## Hello!

## Issue details

The `triggers` property on `local.Command` does not have descriptive text in the docs or in the SDK so it's not clear what is the expected behavior of this property.

screenshot-2022-02-02-12-21-04@2x

thomas11:
Resolves #40
Part of #196 

Issue #40 showed that the `triggers` property could be confusing to users. This PR adds more docs, including sample code.

The actual changes are all in inputs.go and command.md.

Once the general text and approach is approved, I'll extend the docs to the remote Command as well.


provider/pkg/provider/local/command.md
```diff
@@ -0,0 +1,223 @@
...
+```typescript
+import * as local from "@pulumi/command/local";
+import * as random from "@pulumi/random";
+import { asset } from "@pulumi/pulumi";
+import * as path from "path";
+
+const str = "foo";
+const fileAsset = new pulumi.asset.FileAsset("Pulumi.yaml");
+const rand = new random.RandomString("rand", {length: 5});
+const localFile = new command.local.Command("localFile", {
```
corymhall, line 22:
```suggestion
const localFile = new local.Command("localFile", {
```

corymhall (approved):
LGTM!

provider/pkg/provider/local/command.md
```diff
@@ -0,0 +1,223 @@
...
+
+This command can be inserted into the life cycles of other resources using the `dependsOn` or `parent` resource options. A command is considered to have failed when it finished with a non-zero exit code. This will fail the CRUD step of the `Command` resource.
+
+{{% examples %}}
+## Example Usage
+
+{{% example %}}
+### Triggers
+
+This example defines several trigger values of various kinds. Changes to any of them will cause `cmd` to be re-run. However, note that for `fileAsset` it's the variable itself that is the trigger, not the contents of index.ts, since triggers are simply opaque values.
```
danielrbradley, line 11: I'm not quite sure exactly what this is meaning:

> However, note that for `fileAsset` it's the variable itself that is the trigger, not the contents of index.ts, since triggers are simply opaque values

What do you mean by "the variable itself"? I'm guessing this isn't the runtime memory location or anything - do you mean "any property within the object" ... or "it will change every execution" .. or something else?

My assumption was that assets would be checked for changes based on their hash - might be mistaken though.

danielrbradley (approved):
Overall a great improvement to usability!

One area of wording (see comment) is slightly confusing to me.

provider/pkg/provider/local/command.md
thomas11: Yeah it's an awkward phrase. I meant the actual in-memory object of the asset, rather than its content hash. `Triggers` has no special support for assets, it's simply opaque objects.

flostadler (approved):
Nice! Looks good to me, but I agree with Daniel about the `fileAsset` trigger. Could you go into more detail about that one?

provider/pkg/provider/local/command.md
danielrbradley: Ok, I think this is similar to the Random provider's ["keepers" properties](https://www.pulumi.com/registry/packages/random/api-docs/randompassword/#keepers_nodejs)

> Arbitrary map of values that, when changed, will trigger recreation of resource. See the main provider documentation for more information.

However, there's no additional documentation I can find for this provider 🤔 

[The TF docs](https://registry.terraform.io/providers/hashicorp/random/latest/docs#resource-keepers) have the following:

> As noted above, the random resources generate randomness only when they are created; the results produced are stored in the Terraform state and re-used until the inputs change, prompting the resource to be recreated.
> 
> The resources all provide a map argument called keepers that can be populated with arbitrary key/value pairs that should be selected such that they remain the same until new random values are desired.
> 
> Resource "keepers" are optional. The other arguments to each resource must also remain constant in order to retain a random result.
> 
> keepers are not treated as sensitive attributes; a value used for keepers will be displayed in Terraform UI output as plaintext.
> 
> To force a random result to be replaced, the taint command can be used to produce a new result on the next run.

which doesn't help a huge amount either.

Rather than in-memory, perhaps we should be referring to "input value" - i.e. I assume we're waiting for outputs to be resolved and this just behaves in the same way as any other input value. Perhaps the only reason that asset is different is because it's not actually a resource, so doesn't have an "output" property to depend on?

thomas11: Will it even wait for an `Input` to be resolved? I should try it out.

PR #473, opened 2024-06-20
Bring CopyFile back as deprecated resource

thomas11:
To allow users an easy transition to Command 1.0, especially given the lack of alpha and beta releases, we'll keep the previous CopyFile resource for the time being. It will be marked as deprecated, which is currently missing in this PR, therefore it's a draft.

I also tweaked the CopyToRemote integration test for work with CopyFile so we have coverage.


danielrbradley (approved):
LGTM

PR #483, opened 2024-06-28
Add an upgrade-preview test for local commands

thomas11:
As the title says. Recorded via
```
provider $ go test -v github.com/pulumi/pulumi-command/provider/pkg/provider -run TestUpgradeLocalCommand/stdin/upgrade-snapshot -provider-snapshot 
```

It would be good to have one for remote commands as well but that'll take a bit more work. All the integration tests in `examples/` that use remote command also use `CopyToRemote` which doesn't exist in v0.*.


provider/pkg/provider/provider_test.go
```diff
@@ -0,0 +1,31 @@
+// Copyright 2024, Pulumi Corporation.  All rights reserved.
+
+package provider
+
+import (
+	"fmt"
+	"testing"
+
+	"github.com/pulumi/providertest"
```
t0yv0, line 9: I think we may've wanted to deprecate that (@danielrbradley ?) and recommend using providertest/pulumitest for new work.

danielrbradley: ... we should really update the main readme and deprecate the old methods.

thomas11: I rewrote the test using `PreviewProviderUpgrade` and it passes... but stack.json in testdata is also modified, and much shorter, although I didn't re-record??

danielrbradley: I think it records to a slightly different filename IICR so might have just auto-re-recorded first time. No idea why it would be shorter

thomas11: Pushed - could you sanity-check?

danielrbradley: Yup, looks good. Just need to delete the old `state.json` file.

thomas11: Pushed. Ready for a stamp.

danielrbradley (approved):

Issue #433, opened 2024-05-14
Consider adding support for `Diff`


thomas11:
The `Diff` operation is [defined in the Pulumi resource model](https://pulumi-developer-docs.readthedocs.io/en/stable/providers/implementers-guide.html#diff).

> The Diff method is responsible for calculating the differences between the actual and desired state of a resource as represented by its last recorded state and new input Object as returned from [Check](https://pulumi-developer-docs.readthedocs.io/en/stable/providers/implementers-guide.html#check) or [Read](https://pulumi-developer-docs.readthedocs.io/en/stable/providers/implementers-guide.html#read) and the logical operation necessary to reconcile the two (i.e. no operation, an Update, or a Replace).

Allowing users to define a specific command to run on `Diff` could open up new use cases.

This issue is a follow-up from #20.

danielrbradley:
I think this is the same as https://github.com/pulumi/pulumi-command/issues/49 - perhaps we just need a "preview" command which allows the provider to generate the diff.

Issue #432, opened 2024-05-14
Consider adding support for `Read`


thomas11:
The `Read` operation is [defined in the Pulumi resource model](https://pulumi-developer-docs.readthedocs.io/en/stable/providers/implementers-guide.html#read).

> The Read method is responsible for reading the current inputs and state Objects for a resource. Read may be called during a [refresh](https://pulumi-developer-docs.readthedocs.io/en/stable/providers/implementers-guide.html#refresh) or [import](https://pulumi-developer-docs.readthedocs.io/en/stable/providers/implementers-guide.html#import) of a managed resource or during a [preview](https://pulumi-developer-docs.readthedocs.io/en/stable/providers/implementers-guide.html#preview) or [update](https://pulumi-developer-docs.readthedocs.io/en/stable/providers/implementers-guide.html#update) for an external resource.

Allowing users to define a specific command to run on `Read` could open up new use cases.

This issue is a follow-up from #20.

danielrbradley:
Perhaps this should be modelled as "refresh"

vnghia:
Hi do we have any update regarding this issue ? A lot of my resources depend on the output of one `command` provider and they are deleted and re-created everytime I run `pulumi up` :(

If you could give some pointers, I would love to give this feature a try by myself. Pulumi is perfect for me with this feature :D

thomas11:
Hi @vnghia, I'm not quite sure how a custom Read would solve your problem. Could you elaborate? Unless `pulumi up` is run with `--refresh`, it won't call Read.

You might consider changing your command to modify the output, for instance, by piping it to another program.

You could also use [ignoreChanges](https://www.pulumi.com/docs/iac/concepts/options/ignorechanges/) if the output doesn't matter.

vnghia:
I see. Probably something off in my configuration :/ Thank you very much!

Issue #20, opened 2021-12-15
Consider adding support for `diff` and `read`


lukehoban:
These operations are supported in the Resource model, and could be delegated to commands to run on each operation.  The use cases for these are not quite as rich as `create` and `delete`, but could open up additional new use cases.

lukehoban:
The curl example in https://github.com/pulumi/pulumi-command/pull/27 is an example that could in principle benefit from this.  The `color` and `description` of a label can be updated without destroying and recreating the label, but there is no way to represent this today with the API.  The only thing we can implement is that any chance requires replacing the label with a new one.

ghostsquad:
I think the way that go-task does this would work. The provide `status` field, that lets you specify any arbitrary commands to run in order to determine if a "task" should be rerun or not.

https://taskfile.dev/#/usage?id=using-programmatic-checks-to-indicate-a-task-is-up-to-date

The indicator isn't a diff, but rather just the exit code of the `status` commands that are run (if any of them exit with code 1, then the task is _out-of-date_ and needs to rerun).

The `Command` provider could serialize state, and provide it to either `stdin` or as an argument to an `diff` command, which would be able to report whether or not the "resource" would needs to be updated or "recreated".

segator:
we defintelly need update and ideally diff.
I thinking to switch all my company TF projects to pulumi but this is  a big stopper for me, also no support for dynamic provider in GO :(

joeduffy:
Another thing worth considering: `recreateOnUpdate: true`.

For idempotent commands, it would be really nice to trigger re-running the `create` when an update occurs. I've found myself wanting this many times during active development. Otherwise, you need to delete the command, run an update, add back the new command, rerun an update. This can make the inner development cycle a bit laborious.

ossareh:
This would also be helpful in the k8s w/ helm case. I'd like to do something like:

```typescript
const cluster = ...;
const chart = k8s.helm.v3.Release("installation" { ... });
const rollout = new local.Command("rollout", {
    // create not needed because it's the default behavior to start pods when creating the resources in the chart
    update: "kubectl -n namespace rollout deployment installation-foo",
    environment: {
        KUBECONFIG: cluster.kubeconfig,
    },
}, {
    dependsOn: [chart],
});
```

[...]

PR #492, opened 2024-07-11
fix(docs): Update README to include implemented `update` command

stooj:
The README points to https://github.com/pulumi/pulumi-command/issues/20 and says `Update` isn't implemented.

[Diff](https://github.com/pulumi/pulumi-command/issues/433) and [Read](https://github.com/pulumi/pulumi-command/issues/432) are still not done, but [Update is implemented](https://github.com/pulumi/pulumi-command/blob/e7e39fd913708c3e5255d8399c87a736553fec43/sdk/python/pulumi_command/remote/command.py#L196)

The example code could be updated (hah!) to include an `update` (maybe with Luke's `color`/`description` example), but this PR just removes information that misled me.

### DIFF
--- a/README.md
+++ b/README.md
@@ -189,7 +189,7 @@ export const output = rand.stdout;
 
 ### Using `local.Command `with CURL to manage external REST API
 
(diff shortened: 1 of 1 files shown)


thomas11 (approved):
Thank you!

PR #530, opened 2024-09-05
Fix panic on remote connection errors

waketheashes:
There was a panic in `remote.dialWithRetry()` on the context cancellation, e.g. when the timeout specified by `pulumi.Timeouts()` passed to `remote.NewCommand()` exceeded.

`retry.Until()` when the context is cancelled returns: `ok = false, data = nil, err = nil`; so if we try to cast `data` to `T`, we get panic:
> interface conversion: interface {} is nil, not T

We can cast `data` to `T` in `remote.dialWithRetry()`, not only when `err` returned from `retry.Until()` is `nil`, but also when `ok` is `true`.

### DIFF
--- a/provider/pkg/provider/remote/connection.go
+++ b/provider/pkg/provider/remote/connection.go
@@ -122,7 +122,7 @@ func (con *connectionBase) SShConfig() (*ssh.ClientConfig, error) {
 
 func dialWithRetry[T any](ctx context.Context, msg string, maxAttempts int, f func() (T, error)) (T, error) {
 	var userError error
-	_, data, err := retry.Until(ctx, retry.Acceptor{
+	ok, data, err := retry.Until(ctx, retry.Acceptor{
 		Accept: func(try int, _ time.Duration) (bool, any, error) {
 			var result T
 			result, userError = f()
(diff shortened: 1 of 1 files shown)


provider/pkg/provider/remote/connection.go
```diff
@@ -145,11 +145,14 @@ func dialWithRetry[T any](ctx context.Context, msg string, maxAttempts int, f fu
 			return false, nil, nil
 		},
 	})
-	if err == nil {
+	if ok && err == nil {
```
mjeffryes, line 148:
```suggestion
  // It's important to check both `ok` and `err` as sometimes `err` will be need when `ok` is false, such as when the context is cancelled
	if ok && err == nil {
```

provider/pkg/provider/remote/connection.go
```diff
@@ -145,11 +145,14 @@ func dialWithRetry[T any](ctx context.Context, msg string, maxAttempts int, f fu
...
 		return data.(T), nil
 	}
 
 	var t T
+	if err == nil {
+		err = ctx.Err()
```
mjeffryes, line 154:
```suggestion
    // `err` is nil but ok was false, use the err reported from the context.
		err = ctx.Err()
```

mjeffryes (commented):
Thank you for the PR! This looks reasonable to me, though it would be nice to add some comments and a test case. I don't know how difficult it would be to simulate a cancellation in test?

thomas11:
Rebased on latest master and applied comment suggestions in #665

PR #606, opened 2024-12-17
Add docs (copied from pulumi/registry)

iwahbe:
Add docs (copied from `themes/default/content/registry/packages/command/` in pulumi/registry).

This is part of our effort of centralizing information in provider repos.


thomas11 (approved):
I assume the corresponding removal of docs from the registry is tracked separately?

PR #733, opened 2025-04-01
Adopt ci-mgmt.yaml

mjeffryes:
Now that ci-mgmt/provider-ci has a "native" template, we can adopt it for this provider


.github/workflows/artifact-cleanup.yml
```diff
@@ -1,15 +0,0 @@
...
-  - cron: 0 1 * * *
-jobs:
-  remove-old-artifacts:
-    runs-on: ubuntu-latest
-    steps:
-    - name: Remove old artifacts
-      uses: c-hive/gha-remove-artifacts@v1
-      with:
-        age: 1 month
-        skip-tags: true
```
mjeffryes, line 15: We can remove this workflow because we use the expiry in the upload-artifact action instead ([Refactor artifact cleanup · ci-mgmt/484](https://github.com/pulumi/ci-mgmt/pull/484))

.github/workflows/regenerate-sdks.yml
```diff
@@ -1,209 +0,0 @@
...
-  GOVERSION: 1.21.x
-  NODEVERSION: 16.x
-  PYTHONVERSION: "3.7"
-  DOTNETVERSION: |
-    6.0.x
-    3.1.301
-  JAVAVERSION: "11"
-
-jobs:
-  build_provider:
```
mjeffryes, line 15: This workflow seems like it was completely unused and save to remove (no previous runs shown in https://github.com/pulumi/pulumi-command/actions)

EronWright (approved):
LGTM

PR #743, opened 2025-04-23
Fix Makefile

mjeffryes:
I accidentally added the ci-mgmt target before the last line of the sign-goreleaser-exe-% script. Interestingly, this seems to only break the sign-goreleaser-exe-% target so it didn't show up until we attempted to publish!

Fixes: https://github.com/pulumi/pulumi-command/issues/740


Makefile
```diff
@@ -228,4 +229,4 @@ sign-goreleaser-exe-%: bin/jsign-6.0.jar
 ci-mgmt: .ci-mgmt.yaml
 	go run github.com/pulumi/ci-mgmt/provider-ci@master generate
 .PHONY: ci-mgmt
-	fi
+
```
mjeffryes, line 232:
```suggestion
```

PR #747, opened 2025-04-24
Feat: Support SSH Agent named pipe on Windows runner

### LINKED ISSUE #746: Support ssh-agent authentication on Windows
## Hello!

## Issue details

Currently ssh-agent is only supported on Unix: https://github.com/pulumi/pulumi-command/blob/master/provider/pkg/provider/remote/connection.go#L113
To support Windows we'd need to use `winio` to Dial `\\.\pipe\openssh-ssh-agent` named pipe.

ssh-agent is often more convenient to handle the authentication as the key can be used even if it is password protected, and you do not need to put the ssh key password in Pulumi configuration

### Affected area/feature

Should not impact what is currently working, only improve the way ssh connection is handled when the runner is a Windows host

KevinFairise2:
Fixes #746 

On Windows, SSH agent use a named pipe, `\\.\pipe\openssh-ssh-agent` by default.
This PR add some code to try to read that named pipe with `winio` when running on a Windows runner

### DIFF
--- a/provider/pkg/provider/remote/connection.go
+++ b/provider/pkg/provider/remote/connection.go
@@ -18,7 +18,6 @@ import (
 	"context"
 	"fmt"
 	"net"
-	"os"
 	"time"
 
 	p "github.com/pulumi/pulumi-go-provider"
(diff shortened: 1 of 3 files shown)


blampe:
Thank you for putting this together, @KevinFairise2!

Were you able to test this locally, and did it resolve your issue?

Normally I would ask for a quick test, but in this case that will involve spinning up Windows instances on CI, and I don't think this needs to be held up on that.

clarkb7:
@blampe  sorry for the @, is there anything else holding up this PR?

PR #763, opened 2025-05-01
Automated migration to go-provider 1.0

blampe:


provider/pkg/provider/remote/copyController_test.go
```diff
@@ -238,14 +240,15 @@ func TestCheck(t *testing.T) {
 				Archive: archive,
 			},
 		}
-		return resource.NewPropertyMapFromMap(m)
+		pm := resource.NewPropertyMapFromMap(m)
+		return resource.FromResourcePropertyMap(pm)
```
EronWright, line 244: Bit surprising there's not a more direct way that doesn't involve the `resource` package.

EronWright (approved):
LGTM

PR #769, opened 2025-05-04
Add ability to specify host key

### LINKED ISSUE #755: Ability to specify host key
## Hello!

## Issue details

Now the connection completely ignores the host key and does not perform any verification. While this is acceptable for simple one-off commands, for long-living servers it is reasonable to collect the host key, save it and then use to verify the host's identity in future commands.

I propose to add the optional argument "hostKey" to the `Connection` type and use it (if provided) to verify the host key.

https://github.com/pulumi/pulumi-command/blob/4eb4deb77dcb733b1de00d064e00d3ac048f3bbf/provider/pkg/provider/remote/connection.go#L78

Exeteres:
Closes #755

### DIFF
--- a/provider/cmd/pulumi-resource-command/schema.json
+++ b/provider/cmd/pulumi-resource-command/schema.json
@@ -86,6 +86,10 @@
           "type": "string",
           "description": "The address of the resource to connect to."
         },
+        "hostKey": {
+          "type": "string",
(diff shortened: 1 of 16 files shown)


blampe (approved):

blampe:
https://github.com/pulumi/pulumi-command/pull/794

Issue #751, opened 2025-04-28
[docs] The documentation for `command.local.Command()` needs to better explain when commands are executed


aureq:
### What happened?

Reading the documentation for [`command.local.Command()`](https://www.pulumi.com/registry/packages/command/api-docs/local/command/) it may not be clear for everyone that `create` and `update` commands aren't executed at each execution.

The `create` command is only executed **once** when the resource is first created.

The `update` command is executed when at least one of the resource property are changed. But if the code hasn't changed, the `update` command isn't executed on the following run. This behavior may lead to confusion (say deploying a Lambda function) where some expected files aren't present (like in a CI/CD env where `npm ci` is **always** required).

Additionally, the document should point out that `command.local.runOutput()` is available to run specific commands on every execution (including preview).

### Example

In the example below, the function should deploy fine locally on the 1st run ✅. Note that `authorizer/node_modules/` becomes present once deployed since the `create` command has been executed. The file `/tmp/phase.log` confirms this with the line `phase: create`.

Running a 2nd update with no change completes as expected but since there has been no changes to the code, `/tmp/phase.log` isn't updated.

Deleting `authorizer/node_modules` followed by a new deployment fails ❌ when registering the **`aws:lambda/function:Function`** resource. `/tmp/phase.log` isn't updated because the resource `command.local.Command()` hasn't changed.

```
Diagnostics:
  pulumi:pulumi:Stack (zendesk-0016):
    error: Running program '/home/aureq/work/customers/zendesk/0016/index.ts' failed with an unhandled exception:
    <ref *1> Error: failed to register new resource zendesk-lambda-authorizer [aws:lambda/function:Function]: 2 UNKNOWN: failed to compute asset hash for "node_modules/.package-lock.json": failed to open asset file '/home/aureq/work/customers/zendesk/0016/authorizer/node_modules/.package-lock.json': open /home/aureq/work/customers/zendesk/0016/authorizer/node_modules/.package-lock.json: no such file or directory
        at Object.registerResource (/home/aureq/work/customers/zendesk/0016/node_modules/@pulumi/runtime/resource.ts:518:27)
        at new Resource (/home/aureq/work/customers/zendesk/0016/node_modules/@pulumi/resource.ts:555:13)
        at new CustomResource (/home/aureq/work/customers/zendesk/0016/node_modules/@pulumi/resource.ts:1078:9)
[...]

PR #789, opened 2025-05-16
Update docs to clarify when commands are run

blampe:
Fixes https://github.com/pulumi/pulumi-command/issues/751


provider/pkg/provider/common/inputs.go
```diff
@@ -21,10 +21,15 @@ func (c *ResourceInputs) Annotate(a infer.Annotator) {
...
-Command resource from previous create or update steps.`)
+	a.Describe(&c.Create, "The command to run once on resource creation.\n\n"+
+		"If an `update` command isn't provided, then `create` will also be run when the resource's inputs are modified.\n\n"+
+		"Note that this command will not be executed if the resource has already been created and its inputs are unchanged.\n\n"+
+		"Use `local.runOutput` if you need to run a command on every execution of your program.")
+	a.Describe(&c.Delete, "The command to run when the resource is updated.\n\n"+
+		"If empty, the create command will be executed instead.\n\n"+
+		"Note that this command will not run if the resource's inputs are unchanged.\n\n"+
+		"Use `local.runOutput` if you need to run a command on every execution of your program.\n\n"+
+		"The environment variables `PULUMI_COMMAND_STDOUT` and `PULUMI_COMMAND_STDERR` are set to the `stdout` and `stderr` properties of the Command resource from previous create or update steps.")
```
mjeffryes, line 32: I don't think the description here is quite accurate for delete.

```suggestion
	a.Describe(&c.Delete, "The command to run when the resource is deleted.\n\n"+
		"If empty, no command will be run on delete.\n\n"+
		"The environment variables `PULUMI_COMMAND_STDOUT` and `PULUMI_COMMAND_STDERR` are set to the `stdout` and `stderr` properties of the Command resource from previous create or update steps.")
```

provider/pkg/provider/common/inputs.go
```diff
@@ -21,10 +21,15 @@ func (c *ResourceInputs) Annotate(a infer.Annotator) {
...
 	a.Describe(&c.Update, `The command to run on update, if empty, create will
```
mjeffryes, line 33: perhaps you also meant to update the description for update 

```suggestion
	a.Describe(&c.Update, "The command to run when the resource is updated.\n\n"+
		"If empty, the create command will be executed instead.\n\n"+
		"Note that this command will not run if the resource's inputs are unchanged.\n\n"+
		"Use `local.runOutput` if you need to run a command on every execution of your program.\n\n"+
		"The environment variables `PULUMI_COMMAND_STDOUT` and `PULUMI_COMMAND_STDERR` are set to the `stdout` and `stderr` properties of the Command resource from previous create or update steps.")
```

provider/pkg/provider/common/inputs.go
blampe: Ugh thanks good catch!

mjeffryes (approved):

PR #914, opened 2025-09-12
Use ESC for secrets

blampe:
Refs https://github.com/pulumi/ci-mgmt/issues/1481.


guineveresaenger (approved):
Looks like we need to improver our CI output but that's unrelated :/ 
In that case LGTM!

PR #1217, opened 2026-04-23
Fix typo: "delettion" was being used instead of "deletion"

Sodki:


iwahbe (approved):
Thanks for the type fix.

PR #1222, opened 2026-04-27
Support all assets & archives in `CopyToRemote`

iwahbe:
This brings supported values up to match the input types. This is motivated by the ability to use `AssetArchive`s with `CopyToRemote` as part of `pulumi-terraform-converter`.


provider/pkg/provider/remote/copyController.go
```diff
@@ -189,14 +173,174 @@ func copyToRemote(ctx context.Context, input CopyToRemoteInputs) (CopyToRemoteOu
...
+
+	for {
+		name, blob, err := reader.Next()
+		if errors.Is(err, io.EOF) {
+			return nil
+		}
+		if err != nil {
+			return fmt.Errorf("failed to read asset archive entry: %w", err)
+		}
+		if err := writeArchiveEntry(sftpClient, blob, filepath.Join(destPath, name)); err != nil {
```
guineveresaenger, line 308: What if `name` contains `../..` prefixes? a user would have to do that to themselves, but maybe we should guard here to make sure we're still inside `destPath`.

guineveresaenger (approved):
Left a tiny nit but LGTM otherwise.

provider/pkg/provider/remote/copyController.go
iwahbe: If the name climbs out of `destPath`, then it writes above dest path. If the engine lets the user construct such an archive, then the provider should respect it.

If we want to forbid things like this, then IMO that should be applied universally at the engine level.

PR #1231, opened 2026-04-29
Remove "preview" from the README.md

iwahbe:
We are not in preview. We have released v1, and recommend `pulumi-command` to our customers. We will not make breaking changes before v2. We can say as much.


rshade (approved):
LGTM

PR #1351, opened 2026-07-21
EXPERIMENT: Remove Node setup from action.yml

pose:
Removed Node setup step from GitHub Actions workflow.

Part of https://github.com/pulumi/home/issues/4862


pose:
Will experiment on a different way. Resolving!

